PR #32, opened 2017-06-14
Use just context variable

bkeepers:
Merge event and context variable from https://github.com/probot/probot/pull/158

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1,33 +1,28 @@
 const expect = require('expect');
+const createProbot = require('probot');
 const plugin = require('../index');
 
 describe('plugin', () => {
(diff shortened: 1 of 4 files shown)


test/index.js
```diff
@@ -1,33 +1,28 @@
...
-    hooks = {};
-
-    robot = {
-      on: (name, callback) => {
-        hooks[name] = callback;
-      },
-      auth: () => Promise.resolve({})
-    };
+    probot = createProbot({secret: 'test'});
+    probot.robot.auth = () => Promise.resolve({});
```
bkeepers: FIXME: move this into a shared helper

PR #47, opened 2017-09-05
Use standard for linting

hiimbex:
Re: https://github.com/probot/probot/issues/235

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "start": "probot run ./index.js",
-    "test": "mocha && xo"
+    "test": "mocha && standard"
   },
   "author": "Brandon Keepers",
   "license": "ISC",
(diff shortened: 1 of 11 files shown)


hiimbex:
So on this branch for whatever reason I didn't have to add `eslint-plugin-import` re: https://github.com/probot/github-app/pull/9 @gr2m @bkeepers and locally that's still true for me. Just curious if this branch passes tests locally for the both of you?

package.json
```diff
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "start": "probot run ./index.js",
-    "test": "mocha && xo"
+    "test": "mocha && standard"
```
gr2m: I usually put "standard" in a `"pretest"` script, like here:
https://github.com/hoodiehq/hoodie/blob/9bcd805fd01e905dd53b94c5ca547967cdd3f893/package.json#L73

hiimbex: I didn't know npm had a `pretest` field! cool!

gr2m: oh yeah, you can use `pre<script name>` and `post<script name>` for any of the scripts defined in package.json, see https://docs.npmjs.com/misc/scripts

> Pre and post commands with matching names will be run for those as well (e.g. premyscript, myscript, postmyscript)

bkeepers: My preference would be posttest. I would rather focus on getting the tests passing before worrying about the aesthetics of the code.

PR #67, opened 2018-03-13
Extract plugin tests

jwsloan:
To help me understand the code before tackling issue #3, I decided to jump into the tests. It was helpful to split them out, so that I could see how each plugin acts in isolation. 

I wanted to have tests for the `sync` and `update` methods of `Settings.js`, but I couldn't figure out how to get the mocks setup correctly to do so. I wanted those tests to just show the specific duties of those functions, since it is really separate from any one plugin.

I'm not a JS developer, and Google couldn't quite get me far enough. If someone could lead me in the right direction, I'd appreciate it. Or this could be merged, and the other tests left for another PR. It's not my project, so I'll let you all make that call.

/cc @bkeepers


bkeepers:
🎉 Looks great, thanks!

PR #69, opened 2018-03-17
Add milestones configuration

AshanthaLahiru:
Referring to the issue #5

### DIFF
--- a/lib/plugins/milestones.js
+++ b/lib/plugins/milestones.js
@@ -2,7 +2,15 @@ const Diffable = require('./diffable')
 
 module.exports = class Milestones extends Diffable {
   find () {
-    return this.github.issues.getMilestones(this.repo).then(res => res.data)
+    let allMilestones = []
+
+    allMilestones.push(this.github.issues.getMilestones(Object.assign({}, {state: 'open'}, this.repo)).then(res => res.data))
+    allMilestones.push(this.github.issues.getMilestones(Object.assign({}, {state: 'closed'}, this.repo)).then(res => res.data))
+
+    return Promise.all(allMilestones)
+      .then(milestones => {
+        return milestones[0].concat(milestones[1])
+      })
   }
 
   comparator (existing, attrs) {
--- a/lib/plugins/milestones.js
+++ b/lib/plugins/milestones.js
@@ -6,7 +6,9 @@ module.exports = class Milestones extends Diffable {
 
     if (this.entries) {
       this.entries.forEach(milestone => {
-        if (milestone.due_on) { milestone.due_on = new Date(milestone.due_on).toISOString().split('.')[0] + 'Z' }
+        if (milestone.due_on) {
+          delete milestone.due_on
+        }
       })
     }
   }
@@ -20,7 +22,7 @@ module.exports = class Milestones extends Diffable {
   }
 
   changed (existing, attrs) {
-    return existing.description !== attrs.description || existing.state !== attrs.state || existing.due_on !== attrs.due_on
+    return existing.description !== attrs.description || existing.state !== attrs.state
   }
 
   update (existing, attrs) {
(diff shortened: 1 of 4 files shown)


lib/plugins/milestones.js
```diff
@@ -0,0 +1,37 @@
+const Diffable = require('./diffable')
+
+module.exports = class Milestones extends Diffable {
+  find () {
+    let allMilestones = []
+
+    allMilestones.push(this.github.issues.getMilestones(Object.assign({}, {state: 'open'}, this.repo)).then(res => res.data))
+    allMilestones.push(this.github.issues.getMilestones(Object.assign({}, {state: 'closed'}, this.repo)).then(res => res.data))
```
bkeepers: This can be consolidated to one request with `{state: 'all'}`. See https://developer.github.com/v3/issues/milestones/#list-milestones-for-a-repository

lib/plugins/milestones.js
```diff
@@ -0,0 +1,37 @@
...
+
+    return Promise.all(allMilestones)
+      .then(milestones => {
+        return milestones[0].concat(milestones[1])
+      })
```
bkeepers: Switching to one API call should make this go away, but if it didn't, this could be cleaned up by calling `allMilestones.push(...newItems)`. The `...` will expand the array before pushing it onto the new list.

lib/plugins/milestones.js
bkeepers: Also, instead of `then`, most code in Probot uses `async`/`await`:

```js
allMilestones.push((await this.github.issues.getMilestones(params)).data)
```

lib/plugins/milestones.js
```diff
@@ -0,0 +1,37 @@
...
+  comparator (existing, attrs) {
+    return existing.title === attrs.title
+  }
+
+  changed (existing, attrs) {
+    return existing.description !== attrs.description || existing.due_on !== attrs.due_on
+  }
+
+  update (existing, attrs) {
+    return this.github.issues.updateMilestone(Object.assign({}, attrs, this.repo, { number: existing.number }))
```
bkeepers: Minor nit here. You can replace the empty object at the beginning of `Object.assign` with the last argument. In other places an empty object is used as the first arg to create a new object and avoid mutating an existing object that was passed in as a param. But since a new object is being constructed here either way, it can be used.

```diff
- Object.assign({}, attrs, this.repo, { number: existing.number })
+ Object.assign({ number: existing.number }, attrs, this.repo)
```

lib/plugins/milestones.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  add (attrs) {
+    return this.github.issues.createMilestone(Object.assign({}, attrs, this.repo))
+  }
+
+  remove (existing) {
+    return this.github.issues.deleteMilestone(
+      Object.assign({}, { number: existing.number }, this.repo)
```
bkeepers: Same here, you can skip the empty object as the first arg.

AshanthaLahiru:
@bkeepers , thanks a lot for all the clarifications. It was awesome. :smiley:  I will do all the requested changes. :+1:

AshanthaLahiru:
@bkeepers , I did the requested changes. But there is a problem with `due_on` which I missed last times. According to GItHub UI, they only need a date to set Due date for milestone. As I understood even if we give the time also, it will be neglected. But the thing is if I set a date to be `2021-03-23T00:00:00Z`, it will be recorded as `2021-03-22T07:00:00Z`. This will mislead the `due_on` validations. As I feel it's matter of Time Zones, Any thoughts on this? Thanks in advance.

AshanthaLahiru:
@bkeepers , Since GitHub informed `due_on` is an issue of their end, hope we can move forward without it for now until they fix it. Made relevant changes according to that. Hope this will be helpful.

AshanthaLahiru:
Yes

bkeepers:
@AshanthaLahiru Thanks so much for this pull request! Sorry for the delay in getting it merged. It should be available in production now.

Issue #38, opened 2017-07-20
Report errors when parsing settings.yml


bkeepers:
Errors are silently ignored and end-users have no idea why it's not working.

Options:
1. If `.github/settings.yml` is changed in a pull request, a status can be set
2. If it's changed outside of a pull request, open an issue in the repo

danielbachhuber:
I learned this the hard way with https://github.com/wp-cli/widget-command/commit/c39f0f1a6e90427fa2db693c81c54e4e2f658e74

It'd be nice if the YAML document was strictly validated against a spec (although I understand that reporting is the hard part of it).

aryannarora:
@bkeepers In point 1, what is meant by setting a status? Is it like logging or something so the user may know about the changes in `settings.yml` ?

bkeepers:
>  what is meant by setting a status? 

I mean a commit status https://developer.github.com/v3/repos/statuses/

PR #73, opened 2018-03-18
status raise on settings change inside a PR 

aryannarora:
This is in reference to #38.
@bkeepers Do we need something like this?

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,8 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
             owner: repo.owner,
             repo: repo.repo,
             sha: settingChangeCommit.id,
-            state: 'error'
+            state: 'error',
+            context: 'probot/settings'
           })
         } catch (e) {
           console.log(e)
--- a/index.js
+++ b/index.js
@@ -14,12 +14,12 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
       return Settings.sync(context.github, context.repo())
     }
     else if (!defaultBranch && settingsModified) {
-      let settingChangeCommits = payload.commits.filter(commit => {
+      let settingChangeCommits = payload.commits.filter(commit => {  //get all commits that change settings.yml
         return commit.added.includes(Settings.FILE_NAME) || 
         commit.modified.includes(Settings.FILE_NAME)
       })
       
-      settingChangeCommits.forEach(function(commit, index, array) {
+      settingChangeCommits.forEach(function(commit, index, array) { //create status for each commit that changed settings.yml
         context.github.repos.createStatus({
           owner: payload.repository.owner.login,
           repo: payload.repository.name,
--- a/index.js
+++ b/index.js
@@ -2,33 +2,44 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
   robot.on('push', receive)
 
   async function receive (context) {
+    const yaml = require('js-yaml')
+    const fs = require('fs')
     const payload = context.payload
-    const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
+    const defaultBranch =
+      payload.ref === 'refs/heads/' + payload.repository.default_branch
 
     const settingsModified = payload.commits.find(commit => {
-      return commit.added.includes(Settings.FILE_NAME) ||
+      return (
+        commit.added.includes(Settings.FILE_NAME) ||
         commit.modified.includes(Settings.FILE_NAME)
+      )
     })
 
     if (defaultBranch && settingsModified) {
       return Settings.sync(context.github, context.repo())
     } else if (!defaultBranch && settingsModified) {
-      let settingChangeCommits = payload.commits.filter(commit => { // get all commits that change settings.yml
-        return commit.added.includes(Settings.FILE_NAME) ||
-        commit.modified.includes(Settings.FILE_NAME)
-      })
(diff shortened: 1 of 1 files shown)


aryannarora:
@bkeepers Why is it not recognising `createStatus` ? 😢

index.js
```diff
@@ -13,5 +13,20 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     if (defaultBranch && settingsModified) {
       return Settings.sync(context.github, context.repo())
     }
+    else if (!defaultBranch && settingsModified) {
```
knrt10: Remove settingsModified as this condition will not work if it is there

index.js
```diff
@@ -13,5 +13,20 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+      let settingChangeCommits = payload.commits.filter(commit => {  //get all commits that change settings.yml
```
knrt10: instead of using filter here use find() as it will return a single object and you wont have to iterate in array cause you will always get a single file

index.js
aryannarora: @knrt10 Why do you want to remove `settingsModified` ?
We only want to create a status if the settings are modified.
There is no point if we create a status without the settings being changed. :)

index.js
aryannarora: @knrt10 We want to create status for all the commits in a push event that modifies the settings right? Why would there be only one commit in a push event?

index.js
knrt10: Here settings modified will always return undefined if you create changes in any other branch other than master or default branch. You can try and see in console.log. Hope this helps

knrt10: I am sorry I was working on a different repo so my output was not coming. You are right. My bad.

index.js
```diff
@@ -13,5 +13,20 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+        return commit.added.includes(Settings.FILE_NAME) || 
+        commit.modified.includes(Settings.FILE_NAME)
+      })
+      
+      settingChangeCommits.forEach(function(commit, index, array) { //create status for each commit that changed settings.yml
```
knrt10: use ```async function (commit, index, array){```

index.js
```diff
@@ -13,5 +13,20 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+        context.github.repos.createStatus({
```
knrt10: fix it with ```await context.github.repos.createStatus``` . The problem was with promise not it is resolved by it. Try it.

index.js
aryannarora: doing this helps the test to pass but it still says that `createStatus` is undefined. I don't know why. I've used the same thing in my other app and that works fine.

index.js
```diff
@@ -12,6 +12,23 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+        return commit.added.includes(Settings.FILE_NAME) ||
+        commit.modified.includes(Settings.FILE_NAME)
+      })
+      settingChangeCommits.forEach(async function (commit, index, array) { // create status for each commit that changed settings.yml
+        try {
+          await context.github.repos.createStatus({
+            owner: payload.repository.owner.login,
+            repo: payload.repository.name,
+            sha: commit.id,
+            state: 'error'
```
bkeepers: The status of this should only be `error` if the configuration file can't be parsed, or if it contains values that the GitHub API doesn't understand (which is going to be an interesting problem to solve).

index.js
```diff
@@ -12,6 +12,23 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
 
     if (defaultBranch && settingsModified) {
       return Settings.sync(context.github, context.repo())
+    } else if (!defaultBranch && settingsModified) {
+      let settingChangeCommits = payload.commits.filter(commit => { // get all commits that change settings.yml
```
bkeepers: You really only need to set the status on the latest commit in the push. You can do every single one, but most integrations don't bother since GitHub only uses the last one to determine merge-ability.

index.js
```diff
@@ -12,6 +12,23 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+        return commit.added.includes(Settings.FILE_NAME) ||
+        commit.modified.includes(Settings.FILE_NAME)
+      })
+      settingChangeCommits.forEach(async function (commit, index, array) { // create status for each commit that changed settings.yml
+        try {
+          await context.github.repos.createStatus({
```
bkeepers: You'll want to also set a `context` here, which is the name that will show up on GitHub. I would suggest `probot/settings` to mirror the name of this repository.

index.js
```diff
@@ -12,6 +12,23 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+            owner: payload.repository.owner.login,
+            repo: payload.repository.name,
```
bkeepers: You can use `context.repo` for shorthand to extract `owner` and `repo`. See https://probot.github.io/api/latest/Context.html#repo

aryannarora:
@bkeepers Thanks for the review. I'll make the changes as soon as possible. 😄

index.js
aryannarora: @bkeepers Could you please tell a bit more about the second part? I'm curious.

aryannarora:
@bkeepers Could you give me a headstart to check for valid values in `settings.yml`. I'm unable to grasp it.

Issue #1, opened 2017-01-22
Sync settings after installing the integration


bkeepers:
If a repository already has a `.github/config.yml`, then the settings should be synced when the integration is installed.

This involves listening to the [integration webhook events](https://developer.github.com/early-access/integrations/webhooks/) and then looking for a `.github/settings.yml` in all repositories that the installation is given access to.

bkeepers:
It might also be interesting to automatically open a PR that adds all the current settings to `.github/config.yml` if the file doesn't already exist.

knrt10:
@bkeepers I completed this for condition if it is installed in one repository. Should I open a PR so that you can review it?
Thanks in advance.

knrt10:
@bkeepers also regarding the automation to pull request, can you please describe more on this because to open a pull request the user's repo should be forked. Am i right? please correct me and explain.
Thanks in advance

knrt10:
Yes

nesl247:
Yes it is still relevant

pecigonzalo:
@knrt10 your build is not passing.

pecigonzalo:
@knrt10 or @bkeepers

travi:
still relevant

pecigonzalo:
Still relevant @travi could you disable the bot here?

jftanner:
An implementation for this would significantly improve the security stance when using this app.

Copied from my comment on the open PR:
> Currently, an org admin configuring a new repo needs to do two tasks **in order**:
> 1. Install the app in the repo
> 2. Push/merge a commit with `.github/settings.yml` with the required branch protection rules
> 
> If they do those in the wrong order, then it's very easy for them to be under the mistaken impression that the repo has been configured safely. This leaves the repo significantly vulnerable until _either_ the settings are updated or an admin notices the missing rules.
> 
> That condition is especially likely when creating repos from a template, as the template is likely to have the settings file already.

PR #74, opened 2018-03-19
Added enhancement Sync settings after installing the integration

knrt10:
This is reference to issue <a href="https://github.com/probot/settings/issues/1">#1</a>

<strong>Added Functionalities:</strong>

When a user has a predefined ```.github``` folder and they install the probot/settings app after that, then 2 functionalities are added

1.) If ```.github``` folder has ```settings.yml``` file then it will be automatically synced.

2.) If no ```settings.yml``` file is found in ```.github``` then the bot opens a PR  with some default settings in ```settings.yml``` file.

3.) When PR is merged bot automatically delete the branch created by bot.


jwsloan:
Thanks for working on this! Cool idea! 
A concern I have... If I don't like every default setting this provides, then the PR is of no use, correct? What if it actually grabbed the current settings for that repo, and put in a PR creating a matching settings.yml file?

knrt10:
@jwsloan Yea this is nice idea. I will implement this and update the PR.

jwsloan:
Now that I think about it some more, I think this will need to involve creating `importers` for each of the current plugins. The ones that exist now could be moved into an `exporters` file. Then you could look through each setting as it is, and pass those settings to the appropriate importer similar to the way the current plugins work from the settings file.

knrt10:
@jwsloan I implemented that 

> A concern I have... If I don't like every default setting this provides, then the PR is of no use, correct? What if it actually grabbed the current settings for that repo, and put in a PR creating a matching settings.yml file?

But I did'nt get this part 
>Now that I think about it some more, I think this will need to involve creating importers for each of the current plugins. The ones that exist now could be moved into an exporters file. Then you could look through each setting as it is, and pass those settings to the appropriate importer similar to the way the current plugins work from the settings file.

Would you please explain? 
Thanks in advance.

patcon:
It seems like this might be working as submitted. In the interest of supporting a first-time contributor in getting to a successful merge, can we move that import/export architecture suggestion into a new ticket?

jwsloan:
That's a good call, @patcon. It would be great if we could get #76 merged first, and then include the branch protection settings in this. It could lock down master, required Code Owner approval by default.

aryannarora:
@knrt10 Great PR!! :tada: 
 One question though, why did we have to create our own encode function?

knrt10:
@aryannarora Thanks. Whenever we want to create a github file using API our file contents should be `base64` encoded as per [API specifications](https://octokit.github.io/rest.js/#api-Repos-createFile) and also when we receive contents of a repo its encoded so we need to decode it, and I could not find any packages so wrote the function for both encoding and decoding. This PR is using the encoding part and #77 is using both. Hope this helps. :v: .

index.js
```diff
@@ -14,4 +14,91 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+                              '  - name: feature \n' +
+                              '    color: 336699 \n' +
+                              '  - name: first-timers-only \n' +
+                              '    oldname: bug \n\n' +
+                              'collaborators: \n' +
+                              '  - username: enter any username \n' +
+                              '    permission: push'
+
+                const base64 = require('./encode')
+                const encodedString = base64.encode(string)
```
JasonEtco: @knrt10 instead of using some custom created Base64 encoding, you can [use Node.JS's built-in `Buffer` global](https://stackoverflow.com/questions/6182315/how-to-do-base64-encoding-in-node-js):

```js
const string = 'Hello World'
const base64 = Buffer.from(string).toString('base64')
```

index.js
```diff
@@ -14,4 +14,91 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+          } else {
+            try {
+              const reference = await context.github.gitdata.getReference({owner: owner, repo: repoName, ref: 'heads/master'})
+              const refData = reference.data
+              const sha = refData.object.sha
+              try {
+                const getRepo = await context.github.repos.get({owner: owner, repo: repoName})
+                await context.github.gitdata.createReference({owner: owner, repo: repoName, ref: 'refs/heads/probot', sha: sha})
+                // setting the template of file
+                const string = 'repository: \n' +
```
JasonEtco: Can you please move this into a separate file? It'd also be good to use a [template literal]() to avoid all of the `\n` characters. In the separate file, you can do:

```js
module.exports = getRepo => `
  name: ${getRepo.data.name}
`
```

index.js
knrt10: Ok let me update both the changes

index.js
patcon: Ooooo... I am not too proud to admit that I just learned a thing. Thanks! 🎉

index.js
aryannarora: @knrt10 i was asking about this only.

knrt10: @aryannarora finally changed it.

itaditya:
@knrt10 I checked that you are listening to `installation_repositories.added` event for syncing the `.github/settings.yml` file. However this event doesn't trigger when you install the app for the first time. 

So if a user has no `.github/settings.yml` present and installs the app for the first time on his account, then no PR will be made as of now.

knrt10:
@itaditya Thanks. Let me update that too.

itaditya:
for reference you can take a look at https://github.com/itaditya/probot-bootstrap/blob/master/lib/onboard.js#L7

knrt10:
yep still relevant

nesl247:
Any update on this?

knrt10:
yep relevant

knrt10:
Not at all relevant.

patcon:
Just to check in @knrt10 (I'm way out of the loop on the settings plugin, though have recommended it around) -- is this actually not relevant, or was there frustration involved in getting it merged?

PR #76, opened 2018-03-21
Protected branches plugin

### LINKED ISSUE #3: Add support for configuring protected branches
Protected branches have a lot of different toggles: 

- [required status checks](https://developer.github.com/v3/repos/branches/#update-required-status-checks-of-protected-branch)
- [pull request review enforcement](https://developer.github.com/v3/repos/branches/#update-pull-request-review-enforcement-of-protected-branch)
- [team restrictions](https://developer.github.com/v3/repos/branches/#replace-team-restrictions-of-protected-branch)
- [user restrictions](https://developer.github.com/v3/repos/branches/#replace-user-restrictions-of-protected-branch)

I think all of these could be configured via something like this:

```yml
protected_branches:
  master:
    required_status_checks:
      include_admins: true  # Enforce required status checks for repository administrators.
      strict: true # Require branches to be up to date before merging.
      contexts: [continuous-integration/jenkins]
    required_pull_request_reviews: true # boolean or 'non-admin'
    teams: [team1, team2]
    users: [person1, person2]
```

jwsloan:
Fixes #3 .


jwsloan:
@bkeepers I didn't actually test this manually. I couldn't figure out how to get everything setup correctly. My creds were failing. But I figured I'd go ahead and get this submitted anyway.

jwsloan:
@bkeepers I have successfully tested this in a custom app. Any idea when you can take a look?

patcon:
Is your custom app still live? I can do a review, but my JS chops are such that I'd feel best just using it for a week on a non-critical project :)

patcon:
Maybe there's room for a larger discussion on whether we could create two apps (prod and staging), and have a pipeline that deploys to prod on `master` merge. The idea being that @bkeepers can merge master a tiny bit more liberally, knowing that a small crew (maybe technical, but maybe less technical) is on the bleeding edge and will give feedback and/or exercise new features

jwsloan:
@patcon hopefully this works... I just made it public. https://github.com/apps/probotsettingsprotectedbranches

patcon:
Maybe there's room for a larger discussion on whether we could create two apps (prod and staging), and have a pipeline that deploys to prod on `master` merge. The idea being that @bkeepers can merge master a tiny bit more liberally, knowing that a small crew (maybe technical, but maybe less technical) is on the bleeding edge and will give feedback and/or exercise new features

Happy to open another issue if that sounds plausible

patcon:
@jwsloan Playing with it here: https://github.com/patcon/sample-civictech-app

(I'll prob fully exercise it tomorrow -- and it will also help me create a best-practice project starter template for civictech projects for the weekly hacknights I co-organize! so thanks so much!)

README.md
```diff
@@ -88,6 +88,34 @@ teams:
     permission: admin
   - name: docs
     permission: push
+
+branches:
+  - name: 'master'
```
patcon: totes minor, but quoted strings aren't used anywhere else in config :)

README.md
```diff
@@ -88,6 +88,34 @@ teams:
...
+    # https://developer.github.com/v3/repos/branches/#update-branch-protection
+    # Leaving this empty will disable branch protection for this branch.
+    protection:
```
patcon: I'm pretty familiar with this page, and some of my recollection of settings has to do with placement:

screenshot of branch protection config page

Can we put them in the order of the settings screen as much as possible?

jwsloan: We certainly can. That makes a lot of sense. I was ordering it based on the API docs. But basing it on what people are most likely to be familiar with is the right call.

jwsloan:
Thanks for taking a look, @bkeepers. Currently removing the key means you no longer want to manage Protected Branches via this file. It will no longer make any changes. If you leave the `protection` key, but it is empty, it will remove protection.

patcon:
Great point @bkeepers. I like @jwsloan's approach, the removing the key completely should leave protection as-is.

Should we treat `null` and `[]` as empty, just like we do with `{}`? That would seem to be forgiving, even if perhaps not totally consistent. If we do this, would be great to have it reflected in tests :)

jwsloan:
@patcon I went ahead and implemented that. What do you think?

jwsloan:
@patcon, @bkeepers have either of you had a chance to try it out?

patcon:
Yep, I've used it, and it's pretty rad. One thought is that the config is pretty complex, and it might make sense for it to be more forgiving, since probot fails silently.

I know you already extended `null` and `{}`, but what about treating all these the same? `[]`, `{}`, `null`, `false`.

(Happy to defer this into a different issue, as I think it works fine as-if if the person configuring is careful :)

jwsloan:
Thanks for the feedback, @patcon! `[]` should already work. It is definitely reasonable for `false` to be treated as falsey... :) I'll create a test for that.

jwsloan:
@patcon, I've pushed a commit adding tests for `[]` and `false`. They show that it already works as expected. Any other feedback?

jwsloan:
@patcon, sorry. I thought we were just discussing the top-level `protection` key. I'm not sure it makes sense for all of those to be accepted in every situation. I need to look closer at what the API expects.

jwsloan:
@patcon According to [the API documentation](https://developer.github.com/v3/repos/branches/#update-branch-protection), it looks like it is not valid to send `null` for `dismissal_restrictions`. If it is present, it needs to be an object. It seems like this API may not be 100% consistent in how it expects each key to be set to turn it off. 

It seems like an answer is to not have probot fail silently. Is there something we can do about that? It would definitely not be good for someone to think they've successfully protected a branch, but not had it work.

jwsloan:
@bkeepers, @patcon are you guys willing to go ahead and merge this? We could see if people run into pain with the configuration, and go from there?

jwsloan:
@bkeepers the issue this came from just got marked stale. Do you think we could get it merged soon? Thanks!

patcon:
@travi's [comment](https://github.com/probot/settings/issues/3#issuecomment-387268604):

> i see a number of the more complex options for protected branches detailed out above, but my most common use case for simple repos is to simply mark it as protected so no one can accidentally force push. what might that simple configuration look like?

replying here since it has to do with code here :)

My quick impression is that this would work. Perhaps worth having a larger comment code block giving some simple examples though.

```
branches:
  master:
    protection: {}
```

davidjdixon:
@bkeepers I'd like to also see this PR merged. Could we get a yes/no either way?

jwsloan:
Mr. @bkeepers, any chance we can move forward with this?

travi:
thanks @bkeepers, @jwsloan, and everyone for getting this in place. i'm excited to be able to automate this one away

travi:
unfortunately, i'm not seeing this setting take effect. 

i initially tried just setting [what i hoped would be the minimum](https://github.com/travi/scaffolder-sub-command/commit/5b66778e90c4dc338370fe8c0b0cb7fcd6c58be3). when that didn't work, i [paid more attention to the "Required" properties](https://github.com/travi/scaffolder-sub-command/commit/6576e5651cb25197f48d791da74ab2f63dba680e). since that didn't enable protection for the branch either, i tried enabling one of the rules i didnt need, just to see if more was necessary, but still had no luck. am i missing something obvious?

also, just a side note, but the new required properties somewhat conflict with the note under `Usage` that states:
> All settings are optional.

I realize that they are only required if the higher-level settings are provided, but some clarification might be valuable there.

jwsloan:
@travi Since it is live now, I'll give it a shot myself. It could be that the Protected Branches API has changed since I submitted the PR, and I didn't do a good job of keeping track of that.

jwsloan:
@travi would you put in an issue with the settings you're using?

travi:
I can add more detail tonight, but just in my phone for now. The links above show the variations I've tried

travi:
opened a separate issue for what i mentioned above: https://github.com/probot/settings/issues/87

jwsloan:
@patcon does your install of my test app still work? As mentioned in #87, the feature doesn't appear to be working anymore.

PR #82, opened 2018-04-08
Updated Node.js and npm to newer version

aps120797:
Updated Node.js and npm to newer version, supported by Probot


bkeepers:
👍 thanks. This is deployed.

Issue #78, opened 2018-03-30
Label color with leading zeros error


AshanthaLahiru:
If the label color code consists of leading zeros, in some cases it will take it as an integer and it will truncate the leading zeros which leads to following error. 

![settings-label-error](https://user-images.githubusercontent.com/22785263/38132909-a28a34c4-342a-11e8-8967-416d5dfbbace.PNG)

Suggest solution,
![settings-label-solution](https://user-images.githubusercontent.com/22785263/38133756-2f814a40-342e-11e8-88b7-2c970adc0c44.PNG)

AshanthaLahiru:
Found a better solution:

        if (label.color) {
          label.color = String(label.color)
          if(label.color.length < 6) {
            label.color.padStart(6, '0')
          }
        }

PR #85, opened 2018-04-19
Fix label color issue

AshanthaLahiru:
Regarding issue #78


bkeepers:
> This looks good to me. Would you be up for adding a test that demonstrates this?

I pushed a test to verify this.

PR #90, opened 2018-06-26
Add support for probot-config

### LINKED ISSUE #89: add support for probot-config
It would be great to implement support for [probot-config](https://github.com/probot/probot-config). This would enable organization administrators to enforce a number of default repository settings across their org.

pholleran:
This PR fixes #89 and implements support for [probot/config](https://github.com/probot/probot-config)

I have also reformatted `robot.on` within `index.js` to use an arrow function, as per most other probot apps. I've started playing around with changing the `module.exports` statement to be more in line with other probot apps by doing something like:

```js
const getConfig = require('probot-config')
const Settings = require('./lib/settings')

module.exports = robot => {
  robot.on('push', async context => {
    
  <rest_of_function>

  })
}
```

but that started causing test failures I have not yet diagnosed. If that's something that makes sense to add to this PR I'm happy to dig in a bit more.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,8 +5,7 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     const payload = context.payload
     const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
 
-    const config = getConfig(context, 'settings.yml')
-    console.log(config)
+    const config = await getConfig(context, 'settings.yml')
 
     const settingsModified = payload.commits.find(commit => {
       return commit.added.includes(Settings.FILE_NAME) ||
--- a/index.js
+++ b/index.js
@@ -1,7 +1,12 @@
+const getConfig = require('probot-config')
+
 module.exports = (robot, _, Settings = require('./lib/settings')) => {
   robot.on('push', async context => {
     const payload = context.payload
     const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
+
+    const config = getConfig(context, 'settings.yml')
+    console.log(config)
 
     const settingsModified = payload.commits.find(commit => {
       return commit.added.includes(Settings.FILE_NAME) ||
@@ -9,7 +14,7 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     })
 
     if (defaultBranch && settingsModified) {
-      return Settings.sync(context.github, context.repo())
+      return Settings.sync(context.github, context.repo(), config)
     }
   })
 }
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,17 +1,20 @@
...
+const getConfig = require('probot-config')
 
-  async function receive (context) {
+module.exports = (robot, _, Settings = require('./lib/settings')) => {
+  robot.on('push', async context => {
     const payload = context.payload
     const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
 
+    const config = getConfig(context, 'settings.yml')
+    console.log(config)
```
JasonEtco: Can we change this to `context.log()`? Might as well get the free extra data 👌

index.js
```diff
@@ -1,17 +1,20 @@
...
-  robot.on('push', receive)
+const getConfig = require('probot-config')
 
-  async function receive (context) {
+module.exports = (robot, _, Settings = require('./lib/settings')) => {
+  robot.on('push', async context => {
     const payload = context.payload
     const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
 
+    const config = getConfig(context, 'settings.yml')
```
JasonEtco: This needs to be `await`ed otherwise it might return a pending Promise.

index.js
pholleran: @JasonEtco - or we can delete it. That's a remnant of some of my debugging.

JasonEtco:
Oh hm looks like CI is failing now. Let me know if you want help digging into it @pholleran.

pholleran:
@JasonEtco - thanks for offering up some help, I'd appreciate it. I'm pretty new to Jest and am having trouble identifying the underlying problem

JasonEtco:
Hey @pholleran @jwsloan - sorry about the delay, not sure how four days went by but alas here we are.

Of the two (this PR and #93), I think I prefer this approach. It requires less mocking, and these test look clearer to me. I also tend to avoid calling `require` within function/methods whenever possible for performance reasons (negligible I know, but habits are habits).

As for the failing tests, You can see the changes in a61b3c7 - I've created a mocked GitHub API client and made `robot.auth()` return said client. This is more in line with how we typically test Probot Apps (take a look [at our testing docs](https://probot.github.io/docs/testing/), sparse as they are).

If y'all agree I'll go ahead and merge this. Let me know what you think.

pholleran:
@JasonEtco - thanks for the help with the tests. I'm 🆒 with merging this one - just happy to get the functionality added

PR #92, opened 2018-06-29
Disabled required_approving_review_count in README

nesl247:
Commented out `required_approving_review_count` and added a comment on it. The API is currently in preview for this, and thus breaks setting branch protections.

I wasn't sure if I should remove it entirely or comment it out given the comment about the teams above. If it makes more sense to remove it instead, I'm happy to change it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,10 @@ branches:
     protection:
       # Required. Require at least one approving review on a pull request, before merging. Set to null to disable.
       required_pull_request_reviews:
+        # NOTE: The APIs needed for the review count is currently in preview
+        # Uncommenting `required_approving_review_count` will stop the bot from working
         # The number of approvals required. (1-6)
-        required_approving_review_count: 1
+        # required_approving_review_count: 1
         # Dismiss approved reviews automatically when a new commit is pushed.
         dismiss_stale_reviews: true
         # Blocks merge until code owners have reviewed.


nesl247:
Any chance to have this merged?

bkeepers:
Instead of commenting this out, could we add the preview headers to the request?

jwsloan:
@bkeepers, I've got a branch where I have been trying to get all of the preview headers sent that Settings would need, I'm struggling with the Teams headers. Maybe you could help with that? Its probably something easy.

hiimbex:
It appears that API call is being made here: https://github.com/probot/settings/blob/090f010d072a67de75a9c4733a43d083f39aa4f5/lib/plugins/branches.js#L20

and from [these API docs](https://developer.github.com/v3/repos/branches/#update-branch-protection)  it appears the necessary header is `application/vnd.github.luke-cage-preview+json`. So since we use octokit for wrapping the GitHb API, I looked at [their docs on custom headers](https://github.com/octokit/rest.js#api-previews). I think that this line of code: https://github.com/probot/settings/blob/090f010d072a67de75a9c4733a43d083f39aa4f5/lib/plugins/branches.js#L13
could be updated to:
```js
let params = Object.assign(this.repo, { branch: branch.name }, { headers: { accept: 'application/vnd.github.luke-cage-preview+json' }})
```

jwsloan:
Thanks, @hiimbex! Would you have some time to take a look at #98? That is the branch where I'm trying to get all of the preview headers working for this app. I believe I've got the particular one this issue is related to working. I'm struggling with the Teams headers, though.

bkeepers:
Closing since #98 added the required headers to make this work.

PR #93, opened 2018-06-29
Add support for probot-config

jwsloan:
I pulled down PR #90 to help diagnose the test issue. 

While doing so, I thought if we approached the solution a bit differently, it might be easier to test. This is the result of that effort. 

@pholleran, I hope you don't feel like I'm stepping on your toes here. I'd love to hear your feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ node_modules/
 private-key.pem
 .env
 package-lock.json
+.idea/
+probot_settings.iml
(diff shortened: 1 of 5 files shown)


pholleran:
@jwsloan - no toes injured over here. I like your approach - it's more logical.

pholleran:
@JasonEtco - what do you think of this approach?

JasonEtco:
Hey @jwsloan - I just merged #90 for the reasons I described [here](https://github.com/probot/settings/pull/90#issuecomment-401515269). I'll close this out; thanks for giving us another approach!

PR #97, opened 2018-07-19
Upgrade probot and probot-config

jwsloan:


jwsloan:
I didn't notice #96 before creating this. Sorry for the almost-duplicate PR.

jwsloan:
@hiimbex, @gr2m could this get merged?

package.json
```diff
@@ -13,8 +13,9 @@
   "license": "ISC",
   "dependencies": {
     "js-yaml": "^3.7.0",
-    "probot": "^6.2.0",
-    "probot-config": "^0.1.0"
+    "npm": "^6.2.0",
```
gr2m: did you add npm as dependency on purpose?

gr2m: can you remove the change from the PR please?

jwsloan:
@gr2m I've got that change made.

gr2m:
Hmm I don’t get the current error on Travis, any idea?

.travis.yml
```diff
@@ -1,6 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "8"
+  - "10"
```
gr2m: I think I’d keep it to node 8 as far as I know that’s Probot’s currently required version? It will prevent us from using Node 10 features that will break usage in Node 8

gr2m:
The fix looks good! Does it work in Node 8, too? If yes, can you revert the bump in .travis.yml?

jwsloan:
@gr2m done, sir.

PR #100, opened 2018-09-12
WIP: Sync settings from GitHub

### LINKED ISSUE #1: Sync settings after installing the integration

### LINKED ISSUE #41: Sync settings to file on each change
This seems like an extension to https://github.com/probot/settings/issues/1

The proposal would address the concern that enabling this plugin would confuse the process of changing repo settings for people who don't yet know about `probot:settings`. This is perhaps more important for large organizations who might want to introduce the plugin without "breaking" things for people who aren't yet aware.

## Sample painful conversation

**bob:** I've just set up probot:settings!
**alice:** ok great, what happens when someone now makes a change from the UI?
**bob:** good question. I'll assume that we'll have to update the file manually.
**alice:** ok, how will we know when someone changed via UI? And what will be the consequence if we don't notice?
**bob:** good point. We'll have to stay vigilant. If we don't, I assume that the next time the file is changed, it will override the change made via UI. But I'm not sure, I'll have to do some testing, and then we can document the edge-cases.
**alice:** Hm, ok. Given the caveats, is this really worth the process gains?
**bob:** i'm not sure anymore...

## Suggested Feature

To negate these concerns, it would be rad if changes made via the UI were automatically submitted either 1) as updates on default branch, or 2) as a PR to default branch.

(2) seems safer, although might be a little awkward to deal with multiple changes to repo settings, via multiple form saves -- would this involve a specially-named patch branch that is appended to?

Also, there is the outstanding question of whether this requires the addition of a new action to the `RepositoryEvent` type, so that the plugin can know when something has changed:
https://developer.github.com/v3/activity/events/types/#repositoryevent

bkeepers:
This is a _very_ early work in progress to support syncing settings from GitHub. I likely wont have time to work on this any time soon, but wanted to get a PR open with some hacking I did a few days ago.

- [ ] Add a `read` method to fetch current settings for each repository
- [ ] Since each GitHub API response returns a lot of other data, use [octokit/routes](https://github.com/octokit/routes) to determine which values from the response should be saved in the config file
- [ ] Use [probot-scheduler](https://github.com/probot/scheduler) to poll for updated settings for each repository (since there are no webhooks fired for most settings changes)
- [ ] Open a Pull Request whenever settings have changed
- [ ] Open a Pull Request with initial values when the app is installed

Fixes #1
Fixes #41

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,10 @@
 const getConfig = require('probot-config')
 
-module.exports = (robot, _, Settings = require('./lib/settings')) => {
-  robot.on('push', async context => {
+/**
+ * @param {import('probot').Application} app - Probot's Application class.
+ */
+module.exports = async (app, _, Settings = require('./lib/settings')) => {
+  app.on('push', async context => {
     const payload = context.payload
     const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
 
@@ -16,4 +19,13 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
       return Settings.sync(context.github, context.repo(), config)
     }
   })
+
+  const { data: installations } = await (await app.auth()).apps.getInstallations({})
+  const github = await app.auth(installations[0].id)
+  const repos = await github.apps.getInstallationRepositories({})
+
+  const repository = repos.data.repositories[0]
+  const settings = new Settings(github, {owner: repository.owner.login, repo: repository.name}, {})
+
+  console.log('READ', await settings.read())
 }
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -6,6 +6,18 @@ module.exports = class Repository {
     delete this.settings.topics
   }
 
+  async read () {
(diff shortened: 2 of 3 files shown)


travi:
you've probably already considered this, but something to consider either way. what would be the flow for a repo that includes `.github/settings.yml` in the [initial commit](https://github.com/travi/github-scaffolder)?

the flow i'm thinking about would look something like:
* create new repo (empty repo with no initial readme, etc)
* app gets installed automatically since i have it enabled for all repos
* initial commit pushed with `.github/settings.yml` included

would the initial values PR still get created and just need to be closed, or would it not be sent?

Logerfo:
Too bad this is stale. I'd start using the bot if it had this feature.

PR #105, opened 2018-11-14
feat: Add mergeByName

jwsloan:
This is dependent on https://github.com/probot/probot-config/pull/22 getting merged. 

-----
[View rendered README.md](https://github.com/jwsloan/settings/blob/merge-by-name/README.md)

### DIFF
--- a/lib/mergeArrayByName.js
+++ b/lib/mergeArrayByName.js
@@ -1,0 +1,28 @@
+// https://github.com/KyleAMathews/deepmerge#arraymerge
+
+const merge = require('deepmerge')
+
+function findMatchingIndex (sourceItem, target) {
+  if (sourceItem.hasOwnProperty('name')) {
+    return target
+      .filter(targetItem => targetItem.hasOwnProperty('name'))
+      .findIndex(targetItem => sourceItem.name === targetItem.name)
+  }
+}
+
+function mergeByName (target, source, options) {
+  const destination = target.slice()
+
+  source.forEach(sourceItem => {
+    const matchingIndex = findMatchingIndex(sourceItem, target)
+    if (matchingIndex > -1) {
+      destination[matchingIndex] = merge(target[matchingIndex], sourceItem, options)
+    } else {
+      destination.push(sourceItem)
+    }
+  })
+
+  return destination
+}
+
+module.exports = mergeByName
--- a/test/lib/mergeArrayByName.test.js
+++ b/test/lib/mergeArrayByName.test.js
@@ -1,0 +1,76 @@
+const branchArrayMerge = require('../../lib/mergeArrayByName')
+const YAML = require('js-yaml')
+
+describe('mergeArrayByName', () => {
+  it('works', () => {
+    const target = [{
+      name: 'master',
+      shouldChange: 'did not change',
+      shouldKeep: 'kept'
+    }, {
+      name: 'develop'
+    }]
+
+    const source = [{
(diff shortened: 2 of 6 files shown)


lib/mergeArrayByName.js
```diff
@@ -0,0 +1,34 @@
+// https://github.com/KyleAMathews/deepmerge#arraymerge
+
+const merge = require('deepmerge')
+
+function findMatchingIndex (sourceItem, target) {
+  var matchingIndex
+  if (Object.keys(sourceItem).includes('name')) {
+    matchingIndex = target
```
JasonEtco: We can return early here instead of redefining a variable:

```suggestion
    return target
```

lib/mergeArrayByName.js
```diff
@@ -0,0 +1,34 @@
+// https://github.com/KyleAMathews/deepmerge#arraymerge
+
+const merge = require('deepmerge')
+
+function findMatchingIndex (sourceItem, target) {
+  var matchingIndex
+  if (Object.keys(sourceItem).includes('name')) {
```
JasonEtco: Could use `Object.hasOwnProperty` to be just a tiny bit cleaner:

```suggestion
  if (sourceItem.hasOwnProperty('name')) {
```

lib/mergeArrayByName.js
```diff
@@ -0,0 +1,34 @@
...
+    matchingIndex = target
+      .filter((targetItem) => {
```
JasonEtco: Sorry for the nitpick, but I think these would be best as one-liners:

```suggestion
      .filter(targetItem => targetItem.hasOwnProperty('name'))
```

lib/mergeArrayByName.js
```diff
@@ -0,0 +1,34 @@
...
+        return Object.keys(targetItem).includes('name')
+      })
+      .findIndex((targetItem) => {
```
JasonEtco: Same here:

```suggestion
      .findIndex(targetItem => sourceItem.name === targetItem.name)
```

jwsloan:
Thanks for the review, and the tweak to the test, @JasonEtco! I hope you don't mind, I used fixup to combine your tweak into my original commit. I'll do the same with my refactor commit if it is satisfactory.

test/lib/mergeArrayByName.test.js
```diff
@@ -0,0 +1,76 @@
...
+      shouldKeep: 'kept'
+    }, {
+      name: 'develop'
+    }, {
+      name: 'added'
+    }])
+  })
+
+  it('works in a realistic scenario', () => {
+    const target = YAML.load(`
```
JasonEtco: One last thing and then we should be good - internally, everything uses `YAML.safeLoad`, so we should probably use that in our tests too for maximum consistency.

```suggestion
    const target = YAML.safeLoad(`
```

jwsloan:
Thank you, @JasonEtco, for the thoughtful and considerate review! I believe I have satisfied your requests. Also, since Probot was updating anyway, I went ahead and pushed it to 7.4.0.

jwsloan:
I'm sorry, @JasonEtco, but I'm not quite understanding what you mean by end-to-end testing in the case of this app. 

I'm also not certain of the value of such a test. Wouldn't it be testing `probot-config` itself, which is tested on its own? 

Would a test showing that we successfully call `getConfig` with our `mergeArrayByName` function be enough?

PR #138, opened 2019-09-22
Upgrade to node v10

travi:
it looks like the latest LTS version should be supported: https://devcenter.heroku.com/articles/nodejs-support#supported-runtimes

also added an `.nvmrc` to simplify using the intended version locally and on travis. 

in addition, the version of npm bundled with 10.16 is a bit better about churn w/in the lockfile.


.nvmrc
```diff
@@ -0,0 +1 @@
+v10
```
gr2m: We don't have .nvmrc files in other Probot projects and I try to limit the number of dot files we add. I also like that a specific version is defined in the `.travis.yml` file, it's more transparent, less magic. But we can test in node 10 and 12.

I think Probot still supports Node 8 though, why get rid of it? Does that solve a specific problem for you or is it precautionary?

travi: my primary desire for the `.nvmrc` is for local development. in combination with [avn](https://github.com/wbyoung/avn), having an `.nvmrc` in a project is the most convenient way i've found to make sure i'm using the expected node version for a project. this is especially helpful when a project expects a version that is significantly different from what is used by most of the other projects i work on.

i can respect it not being wanted if it diverges from convention though. i'm sure you'll have to nudge me in the preferred direction more than once :)

> I also like that a specific version is defined in the .travis.yml file, it's more transparent, less magic. But we can test in node 10 and 12.

i can understand the desire for being explicit. the reason i tend to simplify is to reduce the risk of it diverging from the `.nvmrc` and `engines` to ensure the ci tests run against the version intended to run in production. 

we certainly can run against multiple versions, but thats probably a question worth me clarifying. can you help me understand the balance between this being an app with a single target production environment and needing to ensure compatibility with multiple versions? would the benefit mostly be for those that run their own instance? i've seen its also not uncommon to publish probot as a versioned package, but it doesnt seem like this one is doing that, at least yet.

> I think Probot still supports Node 8 though, why get rid of it? Does that solve a specific problem for you or is it precautionary?

this probably mostly goes along with the clarification above. my intent isnt necessarily to drop node 8, but upgrade to 10. since only one version was being supported, i assumed we'd stick to that approach, but thought getting to the latest lts would be beneficial. the main benefit i see to at least developing with node 10 is that the bundled version of node introduces less churn in the lockfile. i know lockfiles are highly debated, but i'm on the pro side of that, especially for apps. keeping development on a consistent npm version, especially recent versions, reduces the pain from churn quite a bit.

totally open to sticking with whats best for the project, but that's some context on where my suggestions come from.

gr2m: > can you help me understand the balance between this being an app with a single target production environment and needing to ensure compatibility with multiple versions

That is a very valid point. Let's just test in the same version as we use in production 👍 

Thanks for the thorough explanations.

PR #139, opened 2019-09-24
Upgrade probot to the latest version

travi:


package.json
```diff
@@ -14,7 +14,7 @@
   "dependencies": {
     "deepmerge": "^2.2.1",
     "js-yaml": "^3.12.0",
-    "probot": "^7.4.0",
+    "probot": "^9.5.0",
     "probot-config": "^1.0.0"
```
gr2m: You can remove probot-config now, latest probot has it built-in. That will fix https://github.com/probot/settings/issues/128

travi: good call. thanks for the tip. updated the PR to include that change

PR #140, opened 2019-09-26
Fixed octokit/rest.j sv15.18 deprecations

piccobit:
See here: https://github.com/octokit/rest.js/releases/tag/v15.18.0#deprecations


piccobit:
Sorry for the mess, this is my first contribution.

gr2m:
Not messy at all, thanks for helping out 👍

PR #141, opened 2019-09-28
Added preview 'mercy'

piccobit:
To access the `topics` property for repositories the `Accept` header must contain a custom media type referencing the `mercy-preview`.

### DIFF
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -13,7 +13,10 @@ module.exports = class Repository {
         return this.github.repos.replaceTopics({
           owner: this.settings.owner,
           repo: this.settings.repo,
-          names: this.topics.split(/\s*,\s*/)
+          names: this.topics.split(/\s*,\s*/),
+          mediaType: {
+            previews: ['mercy']
+          }
         })
       }
     })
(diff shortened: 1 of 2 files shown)


lib/plugins/repository.js
```diff
@@ -13,7 +13,10 @@ module.exports = class Repository {
         return this.github.repos.replaceTopics({
           owner: this.settings.owner,
           repo: this.settings.repo,
-          names: this.topics.split(/\s*,\s*/)
+          names: this.topics.split(/\s*,\s*/),
+          mediaType: {
+            previews: ['mercy']
+          }
```
gr2m: Is the preview required to use the endpoint or does it add additional functioality? The note does not say the usual "must set to use this API": https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository, but that might be an error in this case. If so, the preview header should be set by default.

piccobit: Good point! It sounds like you need the `mercy` preview header only to see the `topics` property in the repository results, but from my tests I call tell you that I was not able to set the topics for a repository on our GitHub Enterprise instance.

gr2m: I think it was a mistake, there are other endpoints where the `mercy` preview is optional, such as [Get a repository](https://developer.github.com/v3/repos/#get) (there is a note on top of the page about the mercy preview. I'll file an issue with the docs team. 

But for now, lets get this merged so you get unblocked, it won't hurt to set the preview explicitly

piccobit: Ok, could you then please be so kind and approve this PR?

gr2m:
Sorry @piccobit for making you wait!

PR #143, opened 2019-10-05
Fix label methods from Octokit updates

elstudio:
The current octokit has renamed two label-related methods:

- getLabels has been renamed to listLabelsForRepo
- issues.updateLabel has had two argument changes:
    - oldname attribute has been renamed current_name

This PR addresses both issues, which allows Probot-settings to set up labels once again.


elstudio:
So in the latest, I've wholesale renamed the oldname attribute to current_name, which seems to please octokit and the API.

This means that the oldname attribute should be renamed in settings.yml.

oldname => current_name

This also means that the readme should be updated. I'll can get to that. But does somebody want to see if I'm on the right track here first?

Thanks!

travi:
thanks for the contribution. to help us make sure that make sure that we are on the same page, could you give us a little more background? with the octokit changes, are you seeing warnings or errors as a result? where are you seeing this information? are using the hosted version of the settings app, or hosting your own?

elstudio:
Thanks for asking! In my case, Probot-settings is self-hosted (run from GitHub Actions).

The error kicks in only if there's been a change in labels in settings.yml, which is perhaps why it's not widely reported.

```
14:09:10.559Z ERROR event: this.github.issues.getLabels is not a function (id=1868714c-37cb-4153-9947-9187c1d11b42)
  TypeError: this.github.issues.getLabels is not a function
      at Labels.find (/app/node_modules/probot-settings/lib/plugins/labels.js:22:31)
      at Labels.sync (/app/node_modules/probot-settings/lib/plugins/diffable.js:31:19)
      at Promise.all.Object.entries.map (/app/node_modules/probot-settings/lib/settings.js:24:57)
      at Array.map (<anonymous>)
      at Settings.update (/app/node_modules/probot-settings/lib/settings.js:19:52)
      at github.repos.getContents.then.res (/app/node_modules/probot-settings/lib/settings.js:8:49)
```

elstudio:
Once we replace `this.github.issues.getLabels` with the current `this.github.issues.getLabelsForRepo` method, this error results:

```
23:07:48.004Z DEBUG github: GitHub request: PATCH /repos/:owner/:repo/labels/:current_name - 400
  params: {
    "baseUrl": "https://api.github.com",
    "mediaType": {
      "format": "",
      "previews": []
    },
    "request": {
      "validate": {
        "color": {
          "type": "string"
        },
        "current_name": {
          "required": true,
          "type": "string"
        },
        "description": {
          "type": "string"
           },
        "name": {
          "type": "string"
        },     
        "owner": {
          "required": true,
          "type": "string"
        },
        "repo": {
          "required": true,
          "type": "string"
        }
      },
      "retryCount": 1
    },
    "name": "backlog",
    "color": "ce7048",
    "description": "To be addressed in a future sprint",
    "oldname": "backlog",
    "owner": "elstudio",
    "repo": "reading-time-demo"
  }
{ Deprecation: [@octokit/request-error] `error.code` is deprecated, use `error.status`.
    at RequestError.get (/app/node_modules/@octokit/request-error/dist-node/index.js:29:17)
    at Object.Logger.stdSerializers.err (/app/node_modules/bunyan/lib/bunyan.js:1148:19)
    at /app/node_modules/bunyan/lib/bunyan.js:873:50
    at Array.forEach (<anonymous>)
    at Logger._applySerializers (/app/node_modules/bunyan/lib/bunyan.js:865:35)
    at mkRecord (/app/node_modules/bunyan/lib/bunyan.js:978:17)
    at Logger.error (/app/node_modules/bunyan/lib/bunyan.js:1044:19)
    at Application.<anonymous> (/app/node_modules/probot/lib/application.js:159:33)
    at step (/app/node_modules/probot/lib/application.js:43:23)
    at Object.throw (/app/node_modules/probot/lib/application.js:24:53)
    at rejected (/app/node_modules/probot/lib/application.js:16:65) name: 'Deprecation' }
23:07:48.007Z ERROR event: Empty value for parameter 'current_name': undefined (id=d04acffe-f1f4-41b2-af57-2a95f727c6ad)
  HttpError: Empty value for parameter 'current_name': undefined
      at values.forEach (/app/node_modules/@octokit/rest/plugins/validate/validate.js:74:15)
      at Array.forEach (<anonymous>)
      at Object.keys.forEach.parameterName (/app/node_modules/@octokit/rest/plugins/validate/validate.js:42:12)
      at Array.forEach (<anonymous>)
      at validate (/app/node_modules/@octokit/rest/plugins/validate/validate.js:15:23)
```

gr2m:
The `error.code` deprecation is unrelated and will be resolved with the next Probot version

elstudio:
Indeed!

But the missing current_name attribute that triggered it is the concern. Fixed -- though perhaps not elegantly -- by this PR.

travi:
@elstudio could you help me understand the change from `oldname` to `current_name`? 

i was trying to look through the docs to understand exactly what is meant by that term since the difference between `name` and `current_name` wasn't clear to me. instead of finding that explanation, i instead found mention of [a `new_name` property](https://developer.github.com/v3/issues/labels/#update-a-label).

where did you find information about `current_name`?

gr2m:
I can bring some clarity here. There have been two changes in the `.issues.updateLabel()` method in the past. First, the path for [Update a label](https://developer.github.com/v3/issues/labels/#update-a-label) was `PATCH /repos/:owner/:repo/labels/:name` and the (body) parameter to set a new name was called `name`, too. That URL and body parameters conflicted for usage in Octokit, so I overwrote the URL to `PATCH /repos/:owner/:repo/labels/:oldname`. Then the conflict was resolved in the docs by changing `PATCH /repos/:owner/:repo/labels/:name` to `PATCH /repos/:owner/:repo/labels/:current_name`, hence the deprication.

More recently we started working on an OpenAPI specification for the REST API (see https://github.com/octokit/routes/). There the `PATCH /repos/:owner/:repo/labels/:current_name` route conflicted with `GET /repos/:owner/:repo/labels/:name`. The path mast be the same, so they changed the path back to `PATCH /repos/:owner/:repo/labels/:name` and now accept a new body parameter: `new_name` (the old one still works, is just no longer documented).

I cannot easily create a migration for that because `current_name` was renamed to `name` and `name` was renamed to `new_name`. I will make the change in `@octokit/rest` as part of the next breaking release

gr2m:
tldr: use `current_name` and `name` :)

travi:
thanks @gr2m. thats helpful.

@elstudio could you switch from using the outdated `current_name` to the `new_name` approach? 

also, if you wouldn't mind cleaning up the history of your commits in this PR, i think that would be valuable. i'm not particular about squashing to a single commit for the PR, but cleaning up some of the duplicated commit messages or outdated changes would make the overall changeset more clear.

gr2m:
> @elstudio could you switch from using the outdated current_name to the new_name approach?

Sorry if I wasn't clear. `new_name` will not work with the current version of `probot` / `@octokit/rest`. But the change will be required with the next version, it will be listed in `BREAKING CHANGES` in the release notes.

travi:
> new_name will not work with the current version of probot / @octokit/rest. But the change will be required with the next version

ah, got it. i don't think you were unclear, i just didn't read through carefully enough.

so, it sounds like we should move forward with the current changes and plan to update again once we update the dependencies that include that breaking change.

@elstudio if you could just clean up the commits, i think we can get this merged.

elstudio:
Thanks for the feedback and the forbearance on the messy commits! Commits squashed and comments improved somewhat.

travi:
thanks for getting things cleaned up. i think the changes look good too.

the one piece that i'm a little hesitant about is the `current_name` being a breaking change to anyone with existing config files that attempted to change labels in the past. also because we will need to have another similar breaking change when we move to `new_name`. i think including the breaking changes is appropriate, but we don't currently have a good way to communicate the breaking changes to repository owners that are using the app.

thinking out loud...
* this behavior is currently broken, so old configs already don't work
* old configs that did work at one time would have already had the label modifications applied, so continued processing would be unnecessary
* configs updated to work after this fix would have the modifications applied before the next breaking change, so continued processing would be unnecessary
* while old modifications wouldnt _need_ to be applied, i'm wondering if they would potentially still be attempted and result in an api error, preventing other parts of the config to fail to be applied

@gr2m do you have thoughts around this? it looks like we havent really communicated changes like this in the past and also havent increased the version number either. with your experience around additional probot apps, i'm interested in any insight that you might think is worth considering here.

gr2m:
Ah good catch, I missed that.

I don't think a breaking change is necessary. You could set current name like this

```
attrs.current_name = attrs.current_name || attrs.old_name || attrs.name
```

I would  add a test for the deprecation to make sure it's not accidentally removed.

The settings in the *.yml file don't have to match 1:1 what the octokit methods use, do they? You could just keep using `current_name`, even though in future it will be `name`. I think `current_name` is even better, I just change it because we want to stay aligned with the REST API docs, for better or worse.

travi:
yeah, thats a good call. i really do like the idea of just maintaining backward compatibility since we can here without much trouble.

i do agree that `current_name` is better than just `name`, but i'll admit that the `current` part has been confusing me throughout talking through this PR. without enough context, "current" could be interpreted in multiple ways. 

in addition, i think we should also consider the difference between this being versioned config vs a single api call. by that, i'm referring to the fact that what was "current" before applying the setting change becomes the previous value once the setting is changed. then what is referred to as "current" in the config file is no longer "current". 

personally, i think it would be most clear to use `old_name` and `new_name` to be as clear as possible, regardless of timing. i do think its worth updating to include the underscore for consistency, though.

thoughts?

gr2m:
I find `current_name` more clear than `old_name`, but might be just my personal preference. The "current" states that it is the current name. "old" could refer to a name it had in the past? But at the time of the request, it's still the current name. At least that's my thinking :D

travi:
apologies for being slow to respond again here. i've been trying to think through what made sense, at least to me, for what could work as a good balance between this being a longer-term config and at the time of the update request triggered by the config.

> But at the time of the request, it's still the current name.

i don't disagree with your point here, but i think this is the thing that is tricky in the case of the settings app. the file that triggers the change exists far longer after the change happens than before.

one other option that came to my mind was to use something along the lines of `from_name` and `to_name`. the benefit i see to that option is that it is a little more separated from the time aspect. it doesn't align with the actual api request expectations, but we are already kind of talking about that anyway. if we are already not quite going to match, truly not matching could cut down a little confusion where it almost matches.

one other thing we should keep in mind is the existing use of `name` when not in the process of changing an old name to a new name. this another place where the `current_name`/`new_name` mix feels a little weird. 

admittedly, i haven't had a chance yet to follow through the existing code and understand the mapping fully with the variations of the api that have been discussed, but i wanted to at least get these thoughts back into the conversation. since i'm still ramping up my understanding of the app, i'm really looking for opinions, so please keep any thoughts coming.

gr2m:
> one other thing we should keep in mind is the existing use of name when not in the process of changing an old name to a new name. this another place where the current_name/new_name mix feels a little weird.

I think current_name works well, if you just want to change color or description? But I'm not as familiar with the settings app and the config it provides, I'll let actual users decide :)

travi:
@elstudio since you clearly are using this functionality more than i do (i've only ever provided initial config for labels), does a certain combination of the options mentioned above seem more clear to you?

@jwsloan is this functionality that you've used? do you have feedback in this case?

jwsloan:
I think the code is written to expect the yml to match the API directly. I thought that was pretty clever at the time, but I suppose it is fairly brittle, and doesn't allow for backward compatible changes like you're discussing.

travi:
@elstudio since this PR currently covers two issues, would you mind splitting into two PRs. while we sort out the remaining details for updating labels, i think it makes sense to get the listing update merged. this would also help resolve part of [another report](https://github.com/probot/settings/issues/153) of this issue and a few others.

elstudio:
Sorry @travi, but the tests fail if we fix only the first issue. (The tests also test renaming labels, which of course fail for lack of the current_name attribute. I, too, only use the new labels functionality — but those pesky tests got me into this updateLabel business!)

To your earlier question about yaml property names for `oldname`: What about `renames`? 

Pros: Makes it more clear that if you provide the `renames` property, the code will run updateLabel(), whereas without that property, we do a createLabel().

Cons: `renames` is a verb, so not declarative. Also, the code would have to change a bit to support this, as we can no longer simply pass along any attributes that we’re given. But the new octokit.updateLabel() `name` and `new_name` parameters are going to require this anyway.

What’s the timing for updating the octokit version, which will require `new_name` instead of `current_name`?

lib/plugins/labels.js
```diff
@@ -19,19 +19,19 @@ module.exports = class Labels extends Diffable {
   }
 
   find () {
-    return this.github.issues.getLabels(this.wrapAttrs({})).then(res => res.data)
+    return this.github.issues.listLabelsForRepo(this.wrapAttrs({})).then(res => res.data)
```
kevo1ution: Is it possible to have more than 100 labels, so that pagination is required?

gr2m: Good point. Yes, the code would look like this:

```js
const options = this.github.issues.listLabelsForRepo.endpoint.merge(this.wrapAttrs({per_page: 100}));
return this.github.paginate(options)
```

` this.github.paginate(options)` already returns an array, so no need to map to `res.data`

travi:
> Sorry @travi, but the tests fail if we fix only the first issue. (The tests also test renaming labels, which of course fail for lack of the current_name attribute.

OK, good to know. I wasn't aware of that piece.

> I, too, only use the new labels functionality

also helpful so that i don't lean too hard on you for thoughts on the updating piece :)

> What about renames?

i think that's at least in the realm of what i'm thinking works, so it's a good option to have on the table

> What’s the timing for updating the octokit version, which will require new_name instead of current_name?

i think one of the benefits of the options that somewhat hide the details of the actual api call is that the timing of another change becomes less important. unless we wanted to wait so that we have fewer versions to support, i don't think we need to delay for that reason, especially since it sounds like it sounds like we're currently in a broken state.

gr2m:
> > What’s the timing for updating the octokit version, which will require new_name instead of current_name?

> i think one of the benefits of the options that somewhat hide the details of the actual api call is that the timing of another change becomes less important. unless we wanted to wait so that we have fewer versions to support, i don't think we need to delay for that reason, especially since it sounds like it sounds like we're currently in a broken state.

I want to add that the naming in Octokit is not perfect, it only copies 1:1 whatever the [REST API documentation](https://developer.github.com/v3/) says, for all it's benefits an flaws. By decoupling you from Octokit's names you'll be more future proof and can use clearer names. In this case, the current name of the label will be `name` and the name you want to rename the label to will be `new_name`. I think I would use something more clear than just `name` for the current name, such as `current_name`

travi:
sorry for the delay in bringing this to some sort of resolution. i've been thinking a lot about how we can get a fix in place, since this functionality is currently broken, without growing the complexity too much since the right approach here isn't incredibly clear.

what i think might make the most sense for now is to split the work that is discussed above into a couple of pieces. i think it would be best to simply update the api call to meet the new expectations for now, but without changing how the config file should define the values. i think the work to accept multiple keys, define the recommended one, and clarify that multiple versions could be accepted should be captured in a new issue that can be completed as a separate follow up. that way we can give the second part the appropriate consideration without blocking the fix from being integrated.

@elstudio if you would be willing to update your PR to not change the config file expectations, i would be happy to capture the other details in a new issue.

does this seem like a reasonable compromise to get this moving forward again?

elstudio:
Thanks @travi --

I've updated the PR to leave the config file as it was, with the `oldname` attribute, but to send `current_name` to octokit. 

Not sure how to update the tests to verify the pagination, so I've left that suggestion unaddressed.

kevo1ution:
> Not sure how to update the tests to verify the pagination, so I've left that suggestion unaddressed.

@elstudio Here is an example of testing pagination with listMilestonesForRepo #156

travi:
> Not sure how to update the tests to verify the pagination, so I've left that suggestion unaddressed.

since this was not previously paginated, i think it is ok to merge this PR without handling the pagination quite yet. i've created https://github.com/probot/settings/issues/166 to track that addition separately.

travi:
i've also added https://github.com/probot/settings/issues/167 regarding the backwards compatibility details discussed in https://github.com/probot/settings/pull/143#issuecomment-547227363

travi:
thank you for this contribution @elstudio! sorry for the delay in getting it sorted out and merged, but appreciate the conversation getting to this point. please let me know if this ends up resolving the issues you were seeing originally

Issue #126, opened 2019-06-17
Support notification on error


tunnckoCore:
Seems like my repo doesn't get updated because the #125 where I use `color: #333333` (with the hex), so it errors (probably? the API?)

tunnckoCore:
Seems like it doesn't work at all?

tunnckoCore:
Yup.

gr2m:
can you try `color: "#333333"`? `color: #333333` is valid YML, but everything after # will be ignored as per specification

travi:
it's worth noting that [the api docs for labels](https://developer.github.com/v3/issues/labels/#create-a-label) specifically mention leaving the leading `#` off of the `color`. short of special handling in this application, along the lines of #125, i would only expect this to work without the `#`. also, i do have configs that define colors without the `#` and they are being set correctly using that approach.

tunnckoCore:
Yes, okay, I'm not stupid.

The thing isn't what can be done or that I didn't read the docs. The things is to somehow report such errors so user can know what and why is happening.

Anyway, we will close that.

kyleroush:
The checks API or the status API might be a good way to resolve letting people know if the settings were updated or not.
I think the checks would be the better way as you can provide more information about what failed.

here is a commit where I add it https://github.com/kyleroush/settings/commit/1db507089838db36f2e24f1fd9006fbf0ca3abb5
With this if the settings are all updated it will add a successful check or if the update fails it can describe what failed. Right now when it is failing I am just showing the response body of what failed which is not the prettiest.

gr2m:
I did a similar thing for the WIP app, I use it to explain why a pull request is work in progress. You'd have to add a check in case of a forbidden error, as people are slow to accept new permissions, and you want to handle that case gracefully.

gundalow:
There is another bot (I'll need to find it again) which has a webservice where you can point it at a config file (github.com URL)and it validates it. That might not catch all the problems, though it might help some cases of invalid config.

travi:
i do think linting/validating of a config file could be really useful, but i'm not sure how to do so cheaply with a small team of volunteer maintainers. defining a schema or similar to validate against is the main work that we would need to find time to complete, but i'm open to discussing further, especially if there are ideas around tools that could help simplify. 

this does seem separate from providing feedback as a result of errors, though. would you mind opening a separate issue for conversation around linting/validating of config files? if you do find the tool you're thinking of, i think it would be a valuable detail to include there.

travi:
[...]

PR #148, opened 2019-10-10
support correct hex color

### LINKED ISSUE #125: Support `color` to starts with `#`
Because when we have the hash sign and some plugin on the IDE, you get the color visually.

tunnckoCore:
it's that basic

resolves #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ labels:
     description: An issue with the system 🐛.
 
   - name: feature
-    # If including a `#`, make sure to wrap it with qoutes!
+    # If including a `#`, make sure to wrap it with quotes!
     color: '#336699'
     description: New functionality.
 
--- a/README.md
+++ b/README.md
@@ -63,11 +63,12 @@ labels:
   - name: bug
     color: CC0000
     description: An issue with the system 🐛.
-    
+
   - name: feature
-    color: 336699
+    # If including a `#`, make sure to wrap it with qoutes!
+    color: '#336699'
     description: New functionality.
-    
+
   - name: first-timers-only
     # include the old name to rename an existing label
     oldname: Help Wanted
@@ -131,11 +132,13 @@ branches:
         teams: []
 ```
 
+**Note:** Label color can also starts with `#`, e.g. `color: '#F341B2'`. Make sure to wrap it with qoutes!
+
 ### Inheritance
 
 This app uses [probot-config](https://github.com/probot/probot-config). This means you can inherit settings from another repo, and only override what you want to change.
 
-Individual settings in the arrays listed under `labels`, `teams` (once it is supported) and `branches` will be merged with the base repo if the `name` of an element in the array matches the `name` of an element in the corresponding array in the base repo. A possible future enhancement would be to make that work for the other settings arrays based on `username`, or `title`. This is not currently supported. 
+Individual settings in the arrays listed under `labels`, `teams` (once it is supported) and `branches` will be merged with the base repo if the `name` of an element in the array matches the `name` of an element in the corresponding array in the base repo. A possible future enhancement would be to make that work for the other settings arrays based on `username`, or `title`. This is not currently supported.
 
 To further clarify: Inheritance within the Protected Branches plugin allows you to override specific settings per branch. For example, your `.github` repo may set default protection on the `master` branch. You can then include `master` in your `branches` array, and only override the `required_approving_review_count`.
 Alternatively, you might only have a branch like `develop` in your `branches` array, and would still get `master` protection from your base repo.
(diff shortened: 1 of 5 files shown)


travi:
thanks for the contribution @tunnckoCore. i agree that this would be useful. would you mind also including a test in https://github.com/probot/settings/blob/master/test/lib/plugins/labels.test.js to verify the new behavior (adjusting the one that is there so that it handles both cases would probably be enough too)? with that, i think merging this makes sense.

tunnckoCore:
It be good if so.eone can do that, cuz I'm on phone these days. Otherwise, I may add test when get to pc in the weekend or the next week.

tunnckoCore:
I don't know what's with that bot, or is it slow, or it needs time... but I don't see problems in [this config](https://github.com/tunnckoCore/opensource/blob/348a5b715b3733f57070d60587c01f74d937eb7d/.github/settings.yml)? It probably would be better and faster to type all the 21 labels manually once and for all :D

travi:
> I don't know what's with that bot, or is it slow, or it needs time... but I don't see problems in this config?

I don't understand what this has to do with your PR

> That's why I want some config validation / reporting on problems (#126), we currently only can assume, right?

i understand the desire for improved feedback when an error occurs. thankfully, someone has offered to contribute a PR, which you have found and commented on. again, i don't understand how this comment is useful to this PR.

> regarding the tests, they are not that clear to me what, where and how

i'm happy to help point you in the right direction so we can get this over the finish line. i think it would be enough to update [one of the colors from the simulated config file](https://github.com/probot/settings/blob/aaa60810aecf9fa16426235bccf8363489d7d41a/test/lib/plugins/labels.test.js#L31-L37) to contain a `#` and then ensure that the [corresponding stubbed api](https://github.com/probot/settings/blob/aaa60810aecf9fa16426235bccf8363489d7d41a/test/lib/plugins/labels.test.js#L54-L82) does not contain the `#`.

> Also, it's so natural and intuitive and basic that I don't see the point in adding a test for that. Yea guarantees, but how we can break this anyway...

please keep in mind that this project is now maintained by volunteer labor. personally, i am still learning this codebase, so my confidence is not high, even for small changes. having a test that exercises behavior changes helps give me better confidence that the change won't introduce another case where updates fail in ways that aren't reported to users of the app. there is also regression value so that other changes that happen in the future are safer to introduce without risk of breaking this behavior.

travi:
could you also include a brief note [in the README](https://github.com/probot/settings/blob/aaa60810aecf9fa16426235bccf8363489d7d41a/README.md#L61) that mentions the option of including the `#` so that others can decide if including it would be valuable to them?

tunnckoCore:
Oh, thanks for pointing about tests. Got it.

> could you also include a brief note in the README

For sure.

tunnckoCore:
Done.

README.md
```diff
@@ -131,7 +132,10 @@ branches:
         teams: []
 ```
 
-> **Note**: each top-level element under branch protection must be filled (eg: `required_pull_request_reviews`, `required_status_checks`, `enforce_admins` and `restrictions`). If you don't want to use one of them you must set it to `null` (see comments in the example above). Otherwise, none of the settings will be applied.
+### Notes
+
+1. Label color can also starts with `#`, e.g. `color: '#F341B2'`. Make sure to wrap it with qoutes!
```
travi:
```suggestion
1. Label color can also start with `#`, e.g. `color: '#F341B2'`. Make sure to wrap it with quotes!
```

README.md
```diff
@@ -65,7 +65,8 @@ labels:
     description: An issue with the system 🐛.
 
   - name: feature
-    color: 336699
+    # If including a `#`, make sure to wrap it with qoutes!
```
travi:
```suggestion
    # If including a `#`, make sure to wrap it with quotes!
```

test/lib/plugins/labels.test.js
```diff
@@ -76,7 +76,7 @@ describe('Labels', () => {
           repo: 'test',
           oldname: 'new-description',
           name: 'new-description',
-          color: '000000',
+          color: '#000000',
```
travi: since this is the assertion of what is sent in the api call, we would want the `#` to be removed by this point. i think this will also get the failing build back to passing

```suggestion
          color: '000000',
```

README.md
tunnckoCore: Oh, typos typos :D

README.md
tunnckoCore: Oh yea. :D

tunnckoCore:
Cool. Thanks. :)

travi:
thanks!

PR #149, opened 2019-10-14
Create a check to notify on the outcome of the bot

kyleroush:
This is in response to #126 so the bot can notify when there is an error and provide some additional information.

This would put the check on the last commit that is pushed not on the commits that modified the settings file.

The error in the response indicating what fails will be the json that was returned from the github API.


index.js
```diff
@@ -6,6 +6,13 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     const defaultBranch = payload.ref === 'refs/heads/' + payload.repository.default_branch
 
     const config = await context.config('settings.yml', {}, { arrayMerge: mergeArrayByName })
+    const checkOptions = {
+      owner: context.repo().owner,
+      repo: context.repo().repo,
+      name: "Settings Probot",
```
travi: might be worth using the name of the app, as it appears on the configuration screen

```suggestion
      name: "Settings",
```

travi:
also, it looks like your commits are made using an email address that is not associated with your github account. could you fix that so that your commits can be properly attributed?

kyleroush:
I have added and fixed the test https://github.com/probot/settings/pull/149/files#diff-0fd0e07cf6d02bf7cf00f18cebb8e6ea
I also reset the commits to change who authored them

test/index.test.js
```diff
@@ -27,6 +30,35 @@ describe('plugin', () => {
       await app.receive(event)
       expect(sync).toHaveBeenCalled()
     })
+
+    it('creates a check', async () => {
+      await app.receive(event)
+      expect(github.checks.create).toHaveBeenCalled()
+    })
```
travi: this test seems redundant based on the ones below. if so, could you remove this one?

travi:
thanks for those updates. i think one more minor bit of clean up gets us to where we should be good to move forward.

travi:
@gr2m what would be involved with modifying the requested permissions of the app?

gr2m:
I can make the change. Let me know what you want me to say in the "Add a note to users"

image

travi:
How about

> Settings will now report a status when a commit to the settings file triggers synchronization so that failures can be more easily understood and resolved.

tunnckoCore:
Why not just creating an issue? Then it can be deleted anyway.

> This would put the check on the last commit that is pushed not on the commits that modified the settings file.

And I think exactly the opposite should happen. When settings file is changed **or created for the first time**, it should immediately just open an issue.

test/index.test.js
kyleroush: it did become redundant after i created the more specific test for success and failure 
https://github.com/probot/settings/pull/149/commits/2599ea47fe32a828dca1fc808cf319b4a4f7212b

gr2m:
Check run is the right place to report about a status. Creating issues will required yet more permissions and might create unwanted noise.

travi:
i think this is ready to go. @gr2m how do you want to coordinate the merge and requesting the additional permissions?

jwsloan:
I haven't used the Checks API. What implications are there from having a commit on master unrelated to the settings update have the check attached to it? Or is that what is happening? Maybe I'm not understanding correctly.

kyleroush:
@jwsloan That is what would could happen if they push several commits at once and atleast one of them modified the settings. 

I choose the the latest commit to put the check on rather then the last commit to modify settings because the purpose is to try and notify that settings were updated and i felt the latest one would be the most likely to get noticed for that. 
Also lots of tooling like jenkins/travis tend to run against the last commit and only apply checks/status to the most recent one.

The only real con to this would be navigating through history it might be nice to be able to look at just the settings files and wanting to see which of those commits had the settings applied successfully by looking at the checks on the commits.

But I still feel like the latest commit in the push is the right place to apply the check

travi:
i'm really excited to get this functionality integrated. however, after looking a bit closer i realized that one thing that we haven't handled here is making sure that the permission to set a check run has been granted. handling this would at least improve the roll out of this new functionality so that existing installations aren't left in a broken state until they grant the additional permission.

@kyleroush would you mind adding a check for this? with that in place, i think we'll even be safe to merge ahead of requesting the new permission. that order is probably better anyway so that the promised behavior is available as soon as the requested permission is granted.

travi:
@kyleroush is the additional check that i mentioned something that you would be interested in including in this PR, or would you rather someone else pick that up from here?

index.js
```diff
@@ -14,6 +21,30 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+  ${JSON.stringify(JSON.parse(res.message), null, 2)}
+  </pre>
+</details>
+`
+          checkOptions.output = {
+            title: 'Settings Probot',
+            summary: summary
+          }
+        }).then(() => {
+          context.github.checks.create(checkOptions)
```
kevo1ution: Would it be better to create the check and set it to pending status, and then update it after the settings are synced?

The settings sync can take a lot of time to complete: For example, if they had 50 labels it would take around a minute for the sync to finish but the status check wouldn't exist on their commit until the sync finished.

kyleroush:
:as

sstarcher:
@kyleroush Do you think you might have time  to finish this up or do you want someone else  to jump on and assist?

iphydf:
I'd really like this feature. The PR currently has conflicts. Maybe resolve them and see if everything still passes?

travi:
@kyleroush sorry that i havent followed up further on this. i appreciate you coming back to this and working through the feedback. unfortunately, time has been pretty tight for me lately. this is one of my highest priorities for new features here, but i do think i'm going to take a somewhat different direction for the rollout of this, partly in order to push it out incrementally, but also to be a bit more granular since there are a number of independent things happening for an update. your PR will still be very influential to that solution, so i'm planning to leave this open until i get a chance to work through that. regardless, thank you very much for the contribution and i do plan to get something moving forward when i get the chance.

travi:
> I'd really like this feature. The PR currently has conflicts. Maybe resolve them and see if everything still passes?

many of us would like to see it. there are some complexities to rolling this out in a way that is informative, but also safe. getting this PR back in shape is probably not the best way forward, but it will influence the implementation. unfortunately, time is the main limiting factor in this case, but is high priority.

kyleroush:
So sorry about my poor maintenance of this pr but I have resolved the conflicts with master now

travi:
> So sorry about my poor maintenance of this pr but I have resolved the conflicts with master now

Don't worry. I've been pretty slow on my end as well. I do appreciate you keeping an eye on it.

I've been working to resolve a few other things that were ahead of this, but this feature is now my top priority for this project. I still can't promise much for how quickly I'll have time to work to finalize this, but I do plan to make progress soon.

If you're still up for helping out, which it seems like you are since you got things updated, let's try to keep this PR moving forward.

I pulled down your changes last week to try them out locally and ran into an auth issue with the call to confirm that the app was allowed to create status checks. I didn't dig too far because we were in the middle of upgrading the probot dependency to the next major version. It came with some big changes, so i figured it would be better to investigate further after the upgrade.

the good news is that we are past that upgrade in `master`. would you mind rebasing another time to get your branch in sync with those updates? after that i can start investigating the auth issue further, if it still exists.

i'll likely have a bit of feedback around organization, but lets get past the larger issues first and we can sort out the more minor details afterward.

travi:
i've worked through some of the probot update issues that i mentioned before as well as some auth issues that were unrelated to this PR.

now that i've been able to try this again locally, i can see that the happy path with working as expected and [updates the check to passing](https://github.com/travi-test/settings-test/runs/1046563091). however, upon a failure, the [status remains stuck in the queued state](https://github.com/travi-test/settings-test/runs/1046577633) because an error occurs that ends up not being handled properly. here is the output that appears to be important:

```
ERROR (event): Unexpected token B in JSON at position 0
    SyntaxError: Unexpected token B in JSON at position 0
        at JSON.parse (<anonymous>)
        at /Users/travi/development/probot/settings/index.js:27:27
        at async Promise.all (index 0)
```

could you take a look a look at this issue on your end? i imagine it is something just a bit off with the assumption around the shape of the response. running a local instance will give you the best visibility to the interaction.

thanks for your patience around this functionality. we're close at this point and i'm excited to get this over the finish line

Issue #150, opened 2019-10-16
Branch protection not being applied


renatocaval:
None of the repos where I have branch protection configured gets the settings applied. 

I did two different tests without success.

1. new repo without any rule, rules are not created
1. existing repo with rule for `master`, settings are not applied

Initially, I thought that this could be related to #136. That's why I did test num 2. 

Test were done on this repo: 
https://github.com/renatocaval/oh-my-git/
https://github.com/renatocaval/oh-my-git/blob/master/.github/settings.yml#L13-L23

renatocaval:
You are right. It works if all top-level settings are included. 

Thanks for the tip. 

That's my current one:

```yaml
branches:
  - name: master
    # https://developer.github.com/v3/repos/branches/#update-branch-protection
    # Branch Protection settings. Set to null to disable
    protection:
      # Required. Require at least one approving review on a pull request, before merging. Set to null to disable.
      required_pull_request_reviews:
        # The number of approvals required. (1-6)
        required_approving_review_count: 1
        # Dismiss approved reviews automatically when a new commit is pushed.
        dismiss_stale_reviews: true
      required_status_checks: null
      restrictions: null
      enforce_admins: null
```

I won't work if I set then to `false` like in:

```yaml
branches:
  - name: master
    # https://developer.github.com/v3/repos/branches/#update-branch-protection
    # Branch Protection settings. Set to null to disable
    protection:
      # Required. Require at least one approving review on a pull request, before merging. Set to null to disable.
      required_pull_request_reviews:
        # The number of approvals required. (1-6)
        required_approving_review_count: 1
        # Dismiss approved reviews automatically when a new commit is pushed.
        dismiss_stale_reviews: true
      required_status_checks: false
      restrictions: false
      enforce_admins: false
```

travi:
glad it got you unblocked. thanks for following up with the extra info. that'll be a big help when we get a chance to improve that situation. 

if you'd be interested in sending a PR to improve this, thats always appreciated and i'd be happy to help where you might need it.

renatocaval:
We are now one huge release cycle involving many projects so I don't expect any bandwith before at least end November. Otherwise, I would be willing to help.

On the other hand, with a workaround, it makes it less urgent as well.

travi:
not a problem. i'll leave the issue open until someone can clean that up and so others can hopefully find the workaround in the mean time.

johnmartel:
[...]

PR #151, opened 2019-10-17
Document workaround for branch protection settings

renatocaval:
Relates to #150. It's not a fix, but at least document how to workaround it.

-----
[View rendered README.md](https://github.com/renatocaval/settings/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,21 @@ branches:
         teams: []
 ```
 
+Note: each top-level element under branch protection must be filled (eg: `required_pull_request_reviews`, `required_status_checks`, `enforce_admins` and `restrictions`). If you don't want to use one of them you should set it to `null`. Otherwise, none of the settings will be applied. For example:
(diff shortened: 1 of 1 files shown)


travi:
thanks. this will be very helpful until we can get #150 resolved.

travi:
@renatocaval i'm looking at this a little more closely and thinking about inlining these notes into the larger example in order to be more consistent with existing notes. in looking, i notice that there are already notes about these details being "Required". i could simply include a note for each of those to set to `null` if not wanting to set them, but i'm wondering if they would be noticed. do you think they would have clarified the issue that you were having, or would you have overlooked the note there?

renatocaval:
Hmm, indeed. Maybe we should have just a short heads-up. 

I find it easier to miss the instructions if it's in the form of comments in the config file. As you can see, I didn't read them.

PR #152, opened 2019-10-18
Move note on branch protection to a callout

renatocaval:
* Also remove the previous example. It was redundant                                                                                                
* Points back to the main config example   

The sentence "Otherwise, none of the settings will be applied" is important because it's not obvious that the whole settings will be silently ignored. 

Relates to #151

-----
[View rendered README.md](https://github.com/renatocaval/settings/blob/fix-branch-protection-take2/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,11 @@ labels:
   - name: bug
     color: CC0000
     description: An issue with the system 🐛.
-    
+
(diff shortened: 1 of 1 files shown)


travi:
sorry for the delay here. thank you for the follow-up update

Issue #153, opened 2019-10-21
Errors when updating labels "this.github.issues.getLabels" is not a function


kevo1ution:
I was integrating the settings bot with my own bot, and am getting the error `this.github.issues.getLabels` is not a function in `lib/plugins/labels.js:24.31`. 

I looked at the octokit/rest.js documentation and could not find a function for "issues.getLabels". Should it be `this.github.issues.listLabelsForRepo` (https://octokit.github.io/rest.js/#octokit-routes-issues-list-labels-for-repo) ?

kevo1ution:
I also am having a similar error with milestones: `this.github.issues.getMilestones is not a function lib/plugins/milestones.js:17:42`

travi:
thanks for the report. the `getLabels` issue has been [previously reported](https://github.com/probot/settings/pull/143), but a fix has been slowed as we discuss options around a similar problem with `updateLabel`. i've suggested splitting that PR so that we can get the `getLabels` fix in place. if you have any feedback on either piece of that PR, please feel free to comment there.

i imagine that the `getMilestones` issue that you are seeing is for a similar reason. if you'd be interested in tracking down the details and submitting a PR to fix, i'd be happy to help you through the details to get this resolved.

kevo1ution:
@travi I created a PR for the `getMilestones` issue. I'm not sure if I did it correctly, as this is my first time.

PR #156, opened 2019-10-24
Fixed getMilestones errors and updated api fields

kevo1ution:
Changes:
- converted getMilestones call to use listMilestonesForRepo
- added paginated results for edge cases where there are more than 100 milestones
- updated the api calls to use "milestone_number" instead of "number" field to adhere to the octokit/rest.js [documentation](https://octokit.github.io/rest.js/#octokit-routes-issues-update-milestone)

Issues:
fixes https://github.com/probot/settings/issues/153

### DIFF
--- a/lib/plugins/milestones.js
+++ b/lib/plugins/milestones.js
@@ -14,8 +14,8 @@ module.exports = class Milestones extends Diffable {
   }
 
   async find () {
-    const res = await this.github.issues.getMilestones(Object.assign({ state: 'all' }, this.repo))
-    return res.data
+    const options = this.github.issues.listMilestonesForRepo.endpoint.merge(Object.assign({ state: 'all' }, this.repo))
+    return this.github.paginate(options)
   }
 
   comparator (existing, attrs) {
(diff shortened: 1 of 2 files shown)


travi:
i would prefer not to include the `npm audit fix` commit in this PR. its unrelated enough that it should be a separate PR, if anything. typically though, the maintainers can handle updates like that, so don't feel like its necessary to open a second PR for that.

kevo1ution:
@travi I reverted the commit so that the only changes pertain to fixing the getMilestones issue

lib/plugins/milestones.js
```diff
@@ -14,8 +14,8 @@ module.exports = class Milestones extends Diffable {
   }
 
   async find () {
-    const res = await this.github.issues.getMilestones(Object.assign({ state: 'all' }, this.repo))
-    return res.data
+    const options = this.github.issues.listMilestonesForRepo.endpoint.merge(Object.assign({ state: 'all' }, this.repo))
+    return this.github.paginate(options)
```
travi: shouldn't this still need to wait for an async response?

kevo1ution: Good catch. I actually don't think the function needs to be an async function; I can just return the `this.github.paginate()` promise because in [lib/diffable.js](https://github.com/probot/settings/blob/ef2389b786e339471783142e4d391e5a3efd839f/lib/plugins/diffable.js#L31)

kevo1ution:
@travi I think I fixed the async function issue. Let me know if you want something different.

travi:
thanks for the contribution @kevo1ution. could you please confirm that this resolves the issues you saw and report back?

kevo1ution:
@travi Yes this resolves the issues that I reported

PR #163, opened 2019-11-02
Detect and update config if default branch was changed

mvegter:


travi:
could you explain the use case where this is useful?

we would also want corresponding unit tests to be included if this is a change we end up moving forward with

mvegter:
@travi if the default branches changes than the content of the .github folder could be different. Thus I would expect probot to reload the settings file.

travi:
seems like this should be a really rare use case, but it seems like this is a reasonable way to handle it when it does happen.

with tests in place, i think we can get this merged

mvegter:
Hi @travi, could you have another look. Thanks!

travi:
sorry for the delayed follow up, but thanks a lot for the contribution

PR #165, opened 2019-11-03
Ensure presence of label color

mvegter:


mvegter:
@travi Added the tests, could you have another look.

travi:
since i'm still a relatively new maintainer on this project, i am trying to be extra careful. 

one thing that comes to mind in this case that i'm trying to consider is how the comparator would work in this case. unless i'm missing a detail, i think it would always consider the old and new values to be different, which i would imagine would result in changing to a new random color each time. would you be able to confirm whether this is true?

if this is true, i'm wondering if the complexity of handling that situation is worth the complexity over just requiring the value to be included in the config file. this is essentially the same state as current, but i think with the feedback provided by [a check run (wip)](https://github.com/probot/settings/pull/149) it might be more reasonable to expect this. thoughts about whether that feedback would be enough in this type of case?

mvegter:
The comparator is not affected as it is based on the name of the label. However, your point might be valid for the 'changed' method. 

The web interface of GitHub alreadys generates a random color if you want to add a label, so this behaviour will match using the changes in this PR. I'm not yet sure how generating a random color in the '.github' repo will affect the repos that inherit from this. So I will have to take a look at that too.

PR #170, opened 2019-11-07
Upgrade standard to the latest version

travi:


travi:
@jwsloan it looks like your approval still isnt privileged enough to let me merge. did you end up accepting the invitation?

jwsloan:
I noticed that, @travi . I thought I had, but apparently not. Maybe @gr2m would invite me again.

travi:
assuming that it is still out there, i think you should be able to go to https://github.com/probot/settings/invitations

gr2m:
Should be fixed now! @jwsloan had the Triage permission, I've set it to Write now

travi:
thanks @gr2m

Issue #172, opened 2019-11-11
errors when attempting to use team plugin on latest code  sha - 511b54dafc7a020bd6fb66bd838d0741f1391c45


sstarcher:
The following error is given when attempting to set teams as follows

```teams:
  - name: team1
    permission: admin
  - name: team2
    permission: push
```

``` 19:31:01.734Z ERROR event: Cannot read property 'then' of undefined (id=c92449a6-04b9-11ea-9112-c1b7a775726e)
   TypeError: Cannot read property 'then' of undefined
       at Teams.add (/app/node_modules/probot-settings/lib/plugins/teams.js:25:26)
       at /app/node_modules/probot-settings/lib/plugins/diffable.js:40:31
 19:31:01.734Z ERROR event: Cannot read property 'then' of undefined (id=c92449a6-04b9-11ea-9112-c1b7a775726e)
       at Array.forEach (<anonymous>)
       at /app/node_modules/probot-settings/lib/plugins/diffable.js:34:22
       at async Promise.all (index 2)
   --
   event: {
   TypeError: Cannot read property 'then' of undefined
       at Teams.add (/app/node_modules/probot-settings/lib/plugins/teams.js:25:26)
       at /app/node_modules/probot-settings/lib/plugins/diffable.js:40:31
     "event": "push",
     "id": "c92449a6-04b9-11ea-9112-c1b7a775726e",
       at Array.forEach (<anonymous>)
       at /app/node_modules/probot-settings/lib/plugins/diffable.js:34:22
     "installation": 3511132,
     "repository": "myorg/test"
   }
 19:31:01.735Z ERROR probot: Cannot read property 'then' of undefined
       at async Promise.all (index 2)
   --
   event: {
     "event": "push",
     "id": "c92449a6-04b9-11ea-9112-c1b7a775726e",
   TypeError: Cannot read property 'then' of undefined
       at Teams.add (/app/node_modules/probot-settings/lib/plugins/teams.js:25:26)
     "installation": 3511132,
     "repository": "myorg/test"
   }
 19:31:01.735Z ERROR probot: Cannot read property 'then' of undefined
   TypeError: Cannot read property 'then' of undefined
       at Teams.add (/app/node_modules/probot-settings/lib/plugins/teams.js:25:26)
       at /app/node_modules/probot-settings/lib/plugins/diffable.js:40:31
       at /app/node_modules/probot-settings/lib/plugins/diffable.js:40:31
       at Array.forEach (<anonymous>)
[...]

PR #173, opened 2019-11-13
Fixing pagination in teams plugin

kevo1ution:
Changes
- Instead of using custom pagination, I cleaned it up to use the built-in pagination from Octokit. 

This also fixes the first error found in https://github.com/probot/settings/issues/172


travi:
thanks

PR #178, opened 2019-11-16
Sync settings after App installation

### LINKED ISSUE #1: Sync settings after installing the integration

mvegter:
Fixes #1 

Implementation has overlap with #179

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+const yaml = require('js-yaml')
 const mergeArrayByName = require('./lib/mergeArrayByName')
 
 module.exports = (robot, _, Settings = require('./lib/settings')) => {
@@ -5,6 +6,42 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     const config = await context.config('settings.yml', {}, { arrayMerge: mergeArrayByName })
     return Settings.sync(context.github, repo, config)
   }
+
+  robot.on('installation', async context => {
+    const { github, payload } = context
+    const { action, repositories, installation } = payload
+    const { account } = installation
+    const { login: repositoryOwner } = account
+
+    if (action === 'deleted') {
+      robot.log.debug('Integration deleted, returning...')
+      return
+    }
+
+    await Promise.all(repositories.map(async (repository) => {
+      const { name: repositoryName } = repository
+
+      const repo = {
+        owner: repositoryOwner,
+        repo: repositoryName
+      }
+
+      var res
+      try {
+        res = await github.repos.getContents(Object.assign(repo, { path: Settings.FILE_NAME }))
+      } catch (error) {
+        if (error.status !== 404) {
+          robot.log.warn(`Unknown error ${error.status} occurred when fetching '${Settings.FILE_NAME}' in '${repositoryOwner}/${repositoryName}', returning...`)
+        } else {
+          robot.log.debug(`File '${Settings.FILE_NAME}' not found in '${repositoryOwner}/${repositoryName}', returning...`)
+        }
+        return
+      }
+
+      const config = yaml.safeLoad(Buffer.from(res.data.content, 'base64').toString()) || {}
+      return Settings.sync(context.github, repo, config)
+    }))
+  })
 
   robot.on('push', async context => {
     const { payload } = context
--- a/index.js
+++ b/index.js
@@ -1,4 +1,3 @@
-const yaml = require('js-yaml')
 const mergeArrayByName = require('./lib/mergeArrayByName')
 
 module.exports = (robot, _, Settings = require('./lib/settings')) => {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,6 +1,43 @@
+const yaml = require('js-yaml')
 const mergeArrayByName = require('./lib/mergeArrayByName')
 
 module.exports = (robot, _, Settings = require('./lib/settings')) => {
+  robot.on('installation', async context => {
```
kevo1ution: if you use `installation.added` then you don't need an if statement to check if the action was deleted.

index.js
```diff
@@ -1,6 +1,43 @@
...
+      const repo = {
+        owner: repositoryOwner,
+        repo: repositoryName
+      }
+
+      var res
+      try {
+        res = await github.repos.getContents(Object.assign(repo, { path: Settings.FILE_NAME }))
+      } catch (error) {
+        if (error.status !== 404) {
```
kevo1ution: You should check if the status is `404` otherwise throw the error, so the error can be handled by probot.

index.js
```diff
@@ -1,6 +1,43 @@
...
 const mergeArrayByName = require('./lib/mergeArrayByName')
 
 module.exports = (robot, _, Settings = require('./lib/settings')) => {
+  robot.on('installation', async context => {
+    const { github, payload } = context
+    const { action, repositories, installation } = payload
+    const { account } = installation
+    const { login: repositoryOwner } = account
+
+    if (action === 'deleted') {
```
kevo1ution: This is unnecessary if you use `installation.created` as the [event](https://developer.github.com/v3/activity/events/types/#installationevent ) name. An example of this is shown here with `issues.opened`
https://probot.github.io/docs/hello-world/

index.js
kevo1ution: I believe you also need to include the installations repositories [event](https://developer.github.com/v3/activity/events/types/#installationrepositoriesevent) `installation_repositories.added` which is triggered when a repo is added to the installation. This is a different event from installing a GitHub app which is the `installation.created` event.

kevo1ution: Sorry, I don't think I made this clear: `installation_repositories.added` is still needed as an event to listen for (This event is different from `installation` which is what you currently have). It is triggered when a repo is added to an installation. For example, when you install a bot to your org (that creates an installation and fires `installation` event), and when you add repos to the bot's access later, this fires the `installation_repositories.added` event. 

Also the field name is different to get the repos from the event payload, so watch out for that. https://developer.github.com/v3/activity/events/types/#installationrepositoriesevent

mvegter: @kevo1ution I understood your comment but I decided, for some reason, to seperate the two changes.

index.js
```diff
@@ -1,6 +1,59 @@
...
+      return triggerRepositoryUpdate(context, {
+        owner: repositoryOwner,
+        repo: repositoryName
+      })
+    }))
+  })
+
+  robot.on('installation_repositories.added', async context => {
+    const { payload } = context
+    const { repositories_added: repositories, installation } = payload
```
kevo1ution: It seems that this is very similar to the `installation` event. Have you considered doing something like `repositories = payload.repositories | payload.repositories_added` and just combine the two events.

travi:
am i correct in understanding that this would directly update the file in the default branch? if so, it seems dangerous to do this without submitting the changes as a PR to give the repository maintainer a change to review and accept the changes. if a project already has the settings file, it's likely that there is a reason for it to include the contents that it does, even if it should be supplemented with additional details.

mvegter:
What do you mean with directly updating the settings file?

The flow this results in:
- Probot settings is installed by user (all or selected repos)
- Probot will loop the previous repos and if it finds a settings file it will sync it
- If the user installs additional repos it will repeat this loop for the new repos.

travi:
sorry, i definitely misunderstood the goal. i was thinking about the opposite situation where a repo without an existing file would have one created by pulling the existing settings for that repo through the api and written to the config file. i think there was an issue for that at one point, so i think i made the wrong assumption before reading deeply enough.

with a better understanding now, i do see high value to this. i have been really busy lately, but will try to look through your PR soon and give feedback soon.

travi:
sorry that i still havent gotten to this, but it is still on my radar

ConnectionMaster (approved):

rogerluan:
Maybe ping @travi ? 😊

jftanner:
Is there anything blocking this PR @travi?
This (or another implementation for #1) would significantly improve the security stance when using this app.

Currently, an org admin configuring a new repo needs to do two tasks **in order**:
1. Install the app in the repo
2. Push/merge a commit with `.github/settings.yml` with the required branch protection rules

If they do those in the wrong order, then it's very easy for them to be under the mistaken impression that the repo has been configured safely. This leaves the repo significantly vulnerable until _either_ the settings are updated or an admin notices the missing rules.

That condition is especially likely when creating repos from a template, as the template is likely to have the settings file already.

PR #179, opened 2019-11-16
WIP: If account-level config is changed trigger update all

### LINKED ISSUE #95: Changes in .github repository are not propagated
Sorry, I didn't get a chance to chime in on https://github.com/probot/settings/pull/90, but it doesn't actually work.

As @jwsloan pointed out in https://github.com/probot/probot-config/issues/10#issuecomment-404657030, we'll need to add special support so that if `settings.yml` is updated in a `.github` repository, the changes need to be propagated to all repositories that inherit that config.

cc @pholleran @JasonEtco

mvegter:
Fixes #95

WIP as I the implementation is open for discussion.

flow:
- Update .github repo
- Probot updates .github
- Loop each repo containing settings.yml and update it with the new base config of .github repo

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,91 @@
+const yaml = require('js-yaml')
+const merge = require('deepmerge')
 const mergeArrayByName = require('./lib/mergeArrayByName')
 
 module.exports = (robot, _, Settings = require('./lib/settings')) => {
+  async function onPush (context) {
+    const config = await context.config('settings.yml', {}, { arrayMerge: mergeArrayByName })
+    Settings.sync(context.github, context.repo(), config)
+  }
+
+  async function listRepos (github, { login, type }) {
+    var listForOrg
+    if (type.toLowerCase() === 'organization') {
+      listForOrg = github.paginate(
+        github.repos.listForOrg.endpoint.merge({
+          org: login
+        })
+      )
+    } else {
+      listForOrg = github.paginate(
+        github.repos.listForUser.endpoint.merge({
+          username: login
+        })
+      )
+    }
+
+    return listForOrg
+  }
+
+  async function loadYaml (github, params) {
+    try {
+      const response = await github.repos.getContents(params)
+
+      // Ignore in case path is a folder
+      // - https://developer.github.com/v3/repos/contents/#response-if-content-is-a-directory
+      if (Array.isArray(response.data)) {
+        return null
+      }
+
+      // we don't handle symlinks or submodule
+      // - https://developer.github.com/v3/repos/contents/#response-if-content-is-a-symlink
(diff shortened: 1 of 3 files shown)


gundalow:
Ah, nice. I've just started using Settings Bot, and noticed that none of my changes to github.com/ORG/.github were being propagated to my other repos.

travi:
sorry that it has taken a while to take a dee look at this, but i do really like the approach to this. the alternative of updating the other repos for all commits to the default branch seemed like overkill to me, so i like that this approach is triggered by an actual change to the account-level config instead.

this is a big change to the behavior of this system, so i do want to be careful with getting this integrated. it is also high value, so i definitely want to solve this need.

jay-oswald:
Any updates on this @travi @mvegter would love this to get merged in

travi:
@jay-oswald sorry for the delays, but this is still high on the priority list. there are a few items that i am working to get completed ahead of this in order to make enabling this type of functionality in a safe way. since this is a complex feature, it has the potential to cause problems that we want to avoid.

it has been a little difficult lately to find time to focus on moving this forward safely, but please know that it is still planned to be an upcoming feature

jpoehnelt:
Any support needed to help move this through?

travi:
as mentioned in my recent comment, this is high on the priority list, but needs to be implemented carefully. it is a difficult one to get done through community contributions. please know that it is planned once some other pieces can come together.

index.js
```diff
@@ -21,7 +103,9 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
       return
     }
 
-    const config = await context.config('settings.yml', {}, { arrayMerge: mergeArrayByName })
-    return Settings.sync(context.github, context.repo(), config)
+    await onPush(context)
+    if (repositoryName === '.github') {
```
iphydf, line 107: Do you think we could make this configurable in case I don't want to name my .github repository ".github"? The main reason for that is that I have a [master repo](https://github.com/TokTok/toktok-stack) which has submodules with all the repos in the org. I currently named my "base" repo [template](https://github.com/TokTok/template), but I'm not married to that. However, I can't name it ".github", because then I can't have a submodule for it in the main stack repo, at least not have it named the same way as the .github repo itself. I could have a submodule "template" that points at the repo ".github", but it would be anomalous compared to everything else.

travi: these are the types of details that make this implementation complex. thank you for raising this so that we can consider it. i wasnt personally aware that probot-config supported extending config from other repos than `.github`, but if that is supported there, it probably makes sense to find a way to enable when this goes out. 

would you mind capturing those thoughts in #95 in case this PR ends up not being the way we move this effort forward?

alvarosanchez:
Could this be possibly moved forward?

PR #192, opened 2020-01-31
Bump Node version to 12

JasonEtco:
Removes the `engines` property in `package.json`, because it shouldn't be running Node 10 anymore.

I've removed `check-engine` - I'm not sure that its valuable, but I didn't add it so context would be great! Happy to keep it in place if that's preferred, I can just bump whats in `.engines` instead.


travi:
the goal with `check-engine` is to make sure the local node version (managed with nvm), the ci version (also defaults to nvm), and prod (since heroku will honor the `engines` property in `package.json`) stay in sync. it should be ok to keep node in engines (im fine dropping npm from there) as long as the `.nvmrc` is also updated

PR #193, opened 2020-02-01
docs(issue-templates): added the config.yml

### LINKED ISSUE #174: create issue templates
it would be helpful to ask a few questions up front to have a better idea of the context of where the error was experienced

* using the public instance (currently does not produce error output)
* running own instance against public github
* running own instance against github enterprise

travi:
to get the choices shown when creating a new issue

closes #174


travi:
closing as unnecessary, since adding `about` in #194 did the trick

PR #208, opened 2020-03-05
fix(outside-collaborators): limited the requested list to outside

### LINKED ISSUE #155: modifying .github/settings.yml produces 29 DELETE for collaborators
The project already has all of its collaborators invited, but a single change to the settings produces 29 different instances of

```DEBUG github: GitHub request: DELETE /repos/:owner/:repo/collaborators/:username - 204 No Content (installation=xxx)```

The users are not collaborators of the org they are normal members of the org and the 29 don't seem to have any specific logic to why they have been chosen.  My overall concern is if we expand this to all of our repos we will easily run into API limits.  We currently have hundreds of repos.

travi:
since listing all does not align well with the goals of this configuration. `direct` might also be
reasonable option, but this seems like the most likely fit (see
https://developer.github.com/v3/repos/collaborators/#list-collaborators)

BREAKING CHANGE: repositories relying on comparing to the list of all collaborators will no longer
sync the same way, but that use case seems very unlikely. please open an issue to let us know if you
were relying on this behavior

fixes #155

### DIFF
--- a/lib/plugins/collaborators.js
+++ b/lib/plugins/collaborators.js
@@ -13,17 +13,18 @@ module.exports = class Collaborators extends Diffable {
   }
 
   find () {
-    return this.github.repos.listCollaborators(this.repo).then(res => {
-      return res.data.map(user => {
-        return {
-          // Force all usernames to lowercase to avoid comparison issues.
-          username: user.login.toLowerCase(),
-          permission: (user.permissions.admin && 'admin') ||
-            (user.permissions.push && 'push') ||
-            (user.permissions.pull && 'pull')
-        }
+    return this.github.repos.listCollaborators({ repo: this.repo.repo, owner: this.repo.owner, affiliation: 'direct' })
+      .then(res => {
+        return res.data.map(user => {
+          return {
+            // Force all usernames to lowercase to avoid comparison issues.
+            username: user.login.toLowerCase(),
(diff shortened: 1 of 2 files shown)


travi:
i'm still a little on the fence with this one, so wanted to start with you, @gr2m and @jwsloan in case you happened to know of an api/octokit change that would have impacted this or if you had insight into the original intent of syncing collaborators.

i considered adding a new section for `outside-collaborators` in addition to `collaborators`, but i fail to come up with a situation where the list of _all_ collaborators is the proper list to compare against. my best guess for this not being a bigger problem since the beginning is that removing users that have access through means in addition to being a collaborator had no impact on access, despite the delete requests actually removing them from the collaborators list.

i'm interested in thoughts around if i should consider this more of a breaking change than a bug fix or if we should ping other previous maintainers to get more insight. i certainly lean in the direction of this being a bug fix, but am happy to consider other thoughts.

gr2m:
I don't have any insights on this one I'm afraid :(

travi:
after more thought, i landed on `direct` being more expected than `outside` since it aligns with the web ui. i've updated the PR with that change.

i still can't think of a situation where comparing against all collaborators is expected over `direct`, so i'm comfortable getting this merged if we can get an approval.

PR #209, opened 2020-03-05
fix(teams): migrated away from legacy teams apis

travi:


lib/plugins/teams.js
```diff
@@ -16,29 +16,23 @@ module.exports = class Teams extends Diffable {
...
   }
 
-  toParams (existing, attrs) {
+  toParams (attrs) {
     return {
-      team_id: existing.id,
+      team_slug: attrs.name,
       owner: this.repo.owner,
       repo: this.repo.repo,
+      org: this.repo.owner,
```
travi: it feels wrong for this to be `owner` when `owner` is also provided. its late enough that i'm a bit tired, but am i not considering a situation when this wouldnt be the same?

travi: i think i found the answer to this in reading [the docs](https://developer.github.com/v3/teams/#add-or-update-team-repository) a little more carefully

> The repository must be owned by the organization, or a direct fork of a repository owned by the organization

i didnt realize that an org team could be granted access to a fork, but i guess it somewhat makes sense. 

now i just need to figure out where to get the org from when needing to make this call. i imagine it is available from probot, but need to do some looking to make sure i'm approaching it correctly. if you have this info handy @gr2m, i wouldnt mind a pointer

travi: after thinking about this a bit more, i'm thinking it is valid to assume that the org and the owner are the same unless we were to provide a way to specify an alternative org, whether with an additional attribute or splitting a name specified as `orgName/teamName`.

i'm going to go ahead and merge this for now, but open an issue to handle an alternative org for forks. i'm going to consider that as lower priority unless we get feedback that it is a common need.

PR #221, opened 2020-04-08
fix: return promises in map call to prevent floating promises

leononame:
The branches settings have a subtle error where a map call doesn't actually map into anything but rather creates floating Promises. This PR returns the Promise so that promise rejections can be handled properly in calling code


travi:
thank you for the contribution. could you please add a unit test to verify the behavior of this change?

leononame:
Thanks for the feedback, I added two tests for the promise returns

Issue #225, opened 2020-04-17
Milestones Are not working


maheshrajannan:
## Problem Description

Changing the milestones below in ./github/settings.yml has no effect on either existing milestones or new milestones and there are no new milestones created.

```
# Milestones: define milestones for Issues and Pull Requests
milestones:
  - title: bot-milestone-title
    description: bot-milestone-description
    # The state of the milestone. Either `open` or `closed`
    state: open
```

### What is actually happening

Changing the milestones below in ./github/settings.yml has no effect on either existing milestones or new milestones and there are no new milestones created.

### What is the expected behavior

It should have some effect to start with.

### Error output, if available

```

```

### Are you using the hosted instance of probot/settings or running your own?
Yes.
### If running your own instance, are you using it with github.com or GitHub Enterprise?
Github Enterprise
#### Version of probot/settings
latest.
#### Version of GitHub Enterprise
GitHub Enterprise Server 2.20.0

maheshrajannan:
We are using Github enterprise. and Discover specific information is redacted.
This is what i see in the logs. After seeing this error, i copied the settings.yml from here --> https://github.com/probot/settings/blob/master/README.md

And still see the same issue.

15:01:52.853Z ERROR probot:
--
  | Invalid request.
  |  
  | "branch", "enforce_admins", "required_pull_request_reviews", "required_status_checks", "restrictions" are not permitted keys.
  | --
  | HttpError: Invalid request.
  |  
  | "branch", "enforce_admins", "required_pull_request_reviews", "required_status_checks", "restrictions" are not permitted keys.

travi:
from the error, it looks like the problem is unrelated to milestones, but more likely branch protection. do you have details about branch protection in the same  `settings.yml` file? if there is a failure before the call to configure milestones, the overall run would be aborted, which would explain why you are not seeing the milestones updates that you expect.

travi:
closing since there has been no response to my questions, so i'm assuming the problem has been resolved. please feel free to re-open if there are still issues

thomasklinger1234:
We are encountering the same issue with the default configuration from this repository. Any news?

travi:
thanks for the feedback @thomasklinger1234. i have not been investigating since it appeared above that an unrelated error was preventing the app from getting to the point of applying the milestone changes.

[...]

PR #231, opened 2020-05-08
test(integration): pinned milestones sync behavior

travi:
related to #225


travi:
i plan to split the push test file into separate files soon, but want to get this test added since there is something preventing the syncing from working at runtime

PR #237, opened 2020-05-13
Dependabot

travi:
i think dependabot probably makes the most sense for this particular project for managing updates going forward, but am open to further thoughts. if we do move forward in this direction, i need help actually disabling greenkeeper and enabling dependabot

-----
[View rendered README.md](https://github.com/travi/settings/blob/dependabot/README.md)

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,16 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "live"
+    automerged_updates:
+      - match:
+          dependency_type: "development"
+          update_type: "all"
+      - match:
+          dependency_type: "production"
+          update_type: "all"
+    commit_message:
+      prefix: "fix"
+      prefix_development: "chore"
+      include_scope: true
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # GitHub Settings
 
 [![Build Status](https://img.shields.io/travis/probot/settings.svg)](https://travis-ci.com/probot/settings)
+[![Dependabot][dependabot-badge]][dependabot-link]
 
 This GitHub App syncs repository settings defined in `.github/settings.yml` to GitHub, enabling Pull Requests for repository settings.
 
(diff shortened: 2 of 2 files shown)


.dependabot/config.yml
```diff
@@ -0,0 +1,16 @@
...
+      - match:
+          dependency_type: "development"
+          update_type: "all"
+      - match:
+          dependency_type: "production"
+          update_type: "all"
+    commit_message:
+      prefix: "fix"
+      prefix_development: "chore"
+      include_scope: true
```
gr2m, line 16: There is a new config version (https://dependabot.com/docs/config-file-beta/) that I want to give a try with a config like this:

```yml
version: 2
updates:
  # create PRs for out-of-range updates
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "daily"
    versioning-strategy: "increase-if-necessary"
```

It will only send pull requests for out-of-range updates if I understand correctly, that will drastically reduce the noise

Looks like live updates will be deprecated though, everything will be on schedule 🤷

travi: interesting. i hadn't seen anything about v2 yet. 

> It will only send pull requests for out-of-range updates if I understand correctly, that will drastically reduce the noise

honestly, i'm not personally too concerned about noise as long as they are automerged. i normally define `save-exact=true` in an `.npmrc` for applications (not packages), so out-of-range isn't really an issue in that case. i'm a little surprised i havent configured that here, but probably comes down to thinking that i already had, haha.

i never pushed getting greenkeeper-keeper wired up to this project, but was hoping auto-merge built into dependabot would help in this case. sad to see that it is also on the deprecation list.

gr2m: good point, settings is an app, so we actually do want in-range updates anyway.

> as long as they are automerged

Looks like that is being deprecated with v2 ...

image

travi: yeah, i saw that :(

not as big of an issue with this app since there are so few dependencies, but will certainly influence my usage elsewhere

travi:
thanks for the approval. i think it does make sense to start here at least. we can adjust later if we decide to try out some of the other stuff.

i'll let you merge this once since it should probably be coordinated with the enable/disable of dependabot/greenkeeper that i'll need your help with anyway

gr2m:
> enable/disable of dependabot/greenkeeper

all set 👍

Issue #246, opened 2020-05-15
maintain permissions do not apply on teams


mkrakowitzer:
## Problem Description

maintain permissions do not apply on teams as documented. The API seems to support this behaviour https://developer.github.com/v3/repos/collaborators/

```
teams:
  - name: team1
    permission: maintain
  - name: team2
    permission: maintain
```

The above works on pull, push and admin no problem. If I switch it to maintain it does not apply the change.

### What is actually happening
I assume it's hitting an error because when `maintain` permissions are set as no changes apply to anything in the settings module.

### What is the expected behaviour
Teams should get the maintain permissions

### Error output, if available

```

```

### Are you using the hosted instance of probot/settings or running your own?
Using the hosted version

#### Version of GitHub Enterprise
Enterprise Cloud

mkrakowitzer:
Potentially it is related to this: https://github.com/octokit/octokit.net/issues/2041

The above should be unblocked now as the API Does support both maintain and triage now. 

At the very leas the probot documentation should be updated to state only `pull`, `push`, and `admin` are supported options.

travi:
this _was_ blocked [until very recently](https://github.com/probot/settings/issues/158) because the api did not accept the new values. however, the api for .com was updated to accept them and i verified the update to be working [here](https://github.com/travi-test/settings-test/commit/7a789970daa0368f83d922a246a7a8499ad92e31).

since you are using enterprise, i imagine this is simply a matter of that update not being applied there yet. enterprise usually does get updates _after_ .com, so this should resolve itself once the update makes it that far. i started maintaining settings only fairly recently, so i don't know more specific information about update schedules, but you should be good to go once that rolls out there.

i'm going to close for now since there is no action we can take on the probot/settings side, but feel free to re-open if you think there is more that should be considered

mkrakowitzer:
Hey, Sorry I am using github.com with an enterprise subscription? I was pretty sure this is the same as .com

mkrakowitzer:
@travi I just tested this on a free organisation and it 100% does not work.

travi:
yeah, i'm trying to take a look here too. i agree that it doesnt work from what i'm seeing. trying to understand what i was seeing the other day :\

travi:
we are definitely sending the correct value in the update request, but it's not taking

mkrakowitzer:
@travi No problem, thank you for this great plugin!

travi:
[...]

PR #247, opened 2020-05-15
docs(permissions): removed levels that are unavailable before the oct…

travi:
…okit upgrade

for #246

-----
[View rendered README.md](https://github.com/travi/settings/blob/permissions/README.md)


travi:
i believe these are blocked by the octokit v17 upgrade.

i'll start a draft branch with tests that will fail for now and should pass once these are enabled after the octokit upgrade

PR #248, opened 2020-05-15
New permissions

travi:
adds failing tests for new permissions levels that will be unblocked after probot is upgraded to the latest octokit


travi:
this appears to be resolved with the ocktokit upgrade \o/

Issue #145, opened 2019-10-08
end-to-end tests


travi:
we've had a few issues come up with recent changes that have been related to dependencies that are stubbed by the current isolated unit tests. i think the current tests are still valuable, but also having larger scoped tests that incorporate octokit behavior by exercising calls against the github api (probably stubbed using [nock](https://npm.im/nock)) could help us catch breaking changes that we are currently missing.

a few of the recent things that slipped through:
* https://github.com/probot/settings/pull/143
* https://github.com/probot/settings/pull/140

gr2m:
It's hard to do proper end-to-end tests with Probot right now, definitely an aspect that needs improvement. I'm curious what you come up with, maybe we can adapt Probot's API so make it easier.

nock mocks HTTP requests but does not send webhook event payloads. I always wanted to look into https://github.com/slackapi/steno for that, but didn't have a chance yet

travi:
i was only thinking of using nock where settings makes calls out to the github api, not for the initial webhook that triggers the run. am i thinking wrong that `probot.receive()` is enough for that? i was thinking something along the lines of the [example in the probot docs](https://probot.github.io/docs/testing/).

i'll admit that i haven't gotten through this process yet with a probot app, but it seems like it should be pretty straight forward. now i am curious what i'll run into that you're referring to. 

when i started down this path in the past, the one thing that i remember wishing was possible was to set the http status code that probot would return. beyond that, i dont remember any blockers.

travi:
im hoping to spend some time on this soon. one thing that i've been debating is what to use as the test framework.

jest is obviously being used already on this project, so there could be a benefit to staying consistent with that. i'm not that familiar with jest, but its similar enough to mocha that i'm not concerned about that part.

however, on my own projects i like to use cucumber for larger scoped tests. the main reasons i like this approach is that i think the Given/When/Then interface fits large scoped tests well, but also because it is different from the tools typically used for more isolated unit-tests. I think there is some benefit to mental separation to remind developers that the testing goals are different as they are working with the various test types. 

i recognize that the latter approach is probably more beneficial to longer term contributors than might be the case with an OSS project. i'm not tied to any particular tool, but i'm curious about thoughts on the value of a consistent tool vs. the mental reminder from tools that are targeted a bit differently. i think i'm leaning in the direction of using jest for consistency, but figured i'd ask the question to get more opinions.

@jwsloan @kevo1ution @gr2m thoughts?

gr2m:
[...]

PR #257, opened 2020-06-05
test(integration): split the push test into plugin tests

travi:
for #145

### DIFF
--- a/test/integration/collaborators.test.js
+++ b/test/integration/collaborators.test.js
@@ -1,0 +1,54 @@
+const path = require('path')
+const fs = require('fs')
+const { CREATED, NO_CONTENT, OK } = require('http-status-codes')
+const settings = require('../../lib/settings')
+const { initializeNock, loadInstance, repository, teardownNock } = require('./common')
+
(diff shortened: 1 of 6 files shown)


travi:
as i've added more tests, i've outgrown things being focused on the push event. i'm hoping splitting this way will also make it easier to support additional events that trigger processing

travi:
any idea what is going on with travis for this project @gr2m? looks like its disabled or something: https://travis-ci.com/probot/settings

gr2m:
Hmm I dunno, but maybe let's take the opportunity and migrate tests to actions?

travi:
Yeah, that seems reasonable. Since I don't have visibility to the deployment details, is there anything we need to coordinate there? I think it at least waits for Travis to pass before deploying currently.

gr2m:
> I think it at least waits for Travis to pass before deploying currently

I don't mind it deploying without waiting for tests, as long as we enforce the pull request review process.

travi:
i suppose thats true. since i normally follow trunk-based development, i tend to be a bit more protective of mainline verification

PR #261, opened 2020-06-20
added trigger for syncing based on the repository.created event

travi:
depends on #260. please don't merge before that one


travi:
i decided to go this route afterall because it was a simpler implementation (stays forcused on a single repository for now) and works without needing to listen to additional events (we already listen for `repository.edited`).

PR #271, opened 2020-07-19
fix(package): probot v10

gr2m:
Depends on https://github.com/probot/probot/pull/1250

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1031,9 +1031,9 @@
       }
     },
     "@octokit/graphql": {
-      "version": "4.5.2",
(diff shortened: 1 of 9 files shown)


travi:
@gr2m we're up to date with the latest beta and passing. what are your thoughts around this being merged ahead of you promoting a beta of probot to latest?

gr2m:
One bug was reported to me that I did't yet have the chance to investigate. Once that is done I'd say lets go for it. I'll do the same with WIP

gr2m:
I think this is ready to deploy! Rolling back is simple enough.

PR #385, opened 2020-12-13
feat: Probot v11 (beta 3)

gr2m:


gr2m:
I'm going to merge this and watch the logs as a means to the latest Probot v11 beta release in preparation to release probot v11 stable

gr2m:
looks good now, after #386. I also realized we didn't have the sentry integration setup correctly. I fixed the environment variable, so we should receive errors now

PR #386, opened 2020-12-13
fix: `octokit.repos.replaceTopics` is now `octokit.repos.replaceAllTopics`

gr2m:
we should try to cover more code with integration tests without mocking methods that no longer exist @travi

I went through the unit tests and compared the mocks to the changelogs at https://github.com/octokit/plugin-rest-endpoint-methods.js/releases/v4.0.0, I think we should be good now

### DIFF
--- a/lib/plugins/branches.js
+++ b/lib/plugins/branches.js
@@ -15,7 +15,7 @@ module.exports = class Branches {
           const params = Object.assign(this.repo, { branch: branch.name })
 
           if (this.isEmpty(branch.protection)) {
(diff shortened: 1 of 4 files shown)


travi:
> we should try to cover more code with integration tests without mocking methods that no longer exist @travi

absolutely agree and this is my intent with the integration tests that i've _slowly_ been getting added, but i admittedly dont have integration coverage of all flows yet. thank you for getting these updated. i'll use this as motivation to get coverage in place for at least these sometime soon

PR #387, opened 2020-12-15
test(integration): backfilled coverage for repositories with topics defined

travi:
based on lacking coverage that could have prevented some of the changes needed in #386


gr2m (approved):
thanks, this is great to have!

PR #409, opened 2021-02-18
Make it possible to run probot-settings as a GitHub Action

elstudio:
This PR makes it possible to run probot-settings as a GitHub Action.

If run from a GitHub action, the commit payload lacks the `modified` and `added` properties, see:

https://github.blog/changelog/2019-10-16-changes-in-github-actions-push-event-payload/

This code changes explicitly tests for these properties. If we have them (as if we were run as a typical Probot app), the code logic is unchanged.

But if these properties are missing — as in the case if we are run from an action — then act as if settings.yml has changed.

This makes it possible to run probot-settings from an action like this one: https://github.com/elstudio/actions-settings/pull/11.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,8 +20,15 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     }
 
     const settingsModified = payload.commits.find(commit => {
-      return commit.added.includes(Settings.FILE_NAME) ||
+      // If run from a GitHub action, commits lack these properties, see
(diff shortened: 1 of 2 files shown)


nitrocode:
This would be pretty handy to run as a github action instead of as a bot. An action is much easier to get integrated into an org than an app and can be easier to try it out.

@elstudio would you resolve your conflicts? Maybe it's something that can be merged in?

PR #413, opened 2021-02-21
Fix renaming of labels

### LINKED ISSUE #167: handle various versions of api attributes for label names
this is a follow up for https://github.com/probot/settings/pull/143 fo more forgiving backwards and forwards compatibility with api changes related to label names. as mentioned in https://github.com/probot/settings/pull/143#issuecomment-547227363, deciding a long-term strategy for handling this properly was delaying integration of the fix in that PR, so it was decided to handle this separately.

we may want to handle each of the attribute versions so that all are mapped to the current attribute when the call is made. we may also want to consider a recommended attribute for the config that is more abstracted from the api call and accounts for the timing differences between the config file for the settings app, which has a fairly long life, and the actual api call, which references a specific point in time when the call is made. this timing difference could introduce confusion when trying to understand which term refers to the old value and which refers to the new value (see conversation in https://github.com/probot/settings/pull/143 for additional discussion)

### LINKED ISSUE #407: Labels do not update
## Problem Description

I've added some new labels and renamed a few, but only the new labels are created and the old did not change. 

```yaml
# settings.yml
labels:
  - name: "BC Break"
    color: "#ff0000"
    description: ""
    oldname: "bc break"

  - name: "Bug"
    color: "#fc2929"
    description: ""
    oldname: "bugfix"

  - name: "Enhancement"
    color: "#5fb2d8"
    description: ""
```

### What is actually happening

Existing labels do not change (e.g. rename "bug" to "Bug")

### What is the expected behavior

Label is renamed.

### Are you using the hosted instance of probot/settings or running your own?

Using probot app on github.com 

---

Is there any option to get logs / debug information for the app?

elstudio:
Supports the current API for renaming labels, which expects the following keys in `settings.yml`:

```yaml
labels:
  - name: the old/current name of an existing label we want to change
    new_name: the updated label name
```

Details of this API change are documented here: https://github.com/probot/settings/pull/143#issuecomment-539307068

- Fixes #407 
- Fixes #167 in a dumb way — by passing the attributes needed by the API. The previous `oldname` attribute is no longer supported.

-----
[View rendered README.md](https://github.com/elstudio/probot-settings/blob/rename-labels/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ labels:
     color: '#336699'
     description: New functionality.
 
-  - name: first-timers-only
-    # include the old name to rename an existing label
-    oldname: Help Wanted
+  - name: Help Wanted
+    # Provide a new name to rename an existing label
+    new_name: first-timers-only
 
 # Milestones: define milestones for Issues and Pull Requests
 milestones:
--- a/lib/plugins/labels.js
+++ b/lib/plugins/labels.js
@@ -24,19 +24,22 @@ module.exports = class Labels extends Diffable {
   }
 
   comparator (existing, attrs) {
-    return existing.name === attrs.name || existing.name === attrs.oldname || existing.name === attrs.current_name
+    return existing.name === attrs.name
   }
 
   changed (existing, attrs) {
-    return attrs.oldname === existing.name || existing.color !== attrs.color || existing.description !== attrs.description
+    return 'new_name' in attrs || existing.color !== attrs.color || existing.description !== attrs.description
   }
 
   update (existing, attrs) {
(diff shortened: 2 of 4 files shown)


travi:
thank you for the contribution. apologies for the delay in responding on this.

i did initially want to handle attribute changes like this so that org level changes within a `.github` repository wouldnt be left broken over time since they don't get applied immediately (which is planned to be improved once i find the time). however, since this has been broken now for a while, i do think it is appropriate to support only the current format. backwards compatibility support could come at a later time.

travi (approved):

PR #473, opened 2021-07-05
ci(workflow): use npm ci in node-cli.yml

### LINKED ISSUE #472: node-ci cache and GitHub Action
After adding native `cache` support for `setup-node` action I have some doubts on this step:
https://github.com/probot/settings/blob/b0ca60a3a6f8e498e0c7cf98742707f68c3ada45/.github/workflows/node-ci.yml#L21

Is it adding a cache so it can be replaced by a regular `npm install` or `npm ci` ? Or is there anything else to keep this step (and maybe rollback `cache` feature from `setup-node` step)?

oscard0m:
Fix #472

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -18,5 +18,5 @@ jobs:
         with:
           node-version: 12.x
           cache: npm
-      - uses: bahmutov/npm-install@v1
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thanks!

PR #515, opened 2021-11-10
🚧 Vercel deployment

gr2m:


gr2m:
Well that was absolutely painless :) @travi 

I created https://github.com/apps/gr2m-settings/ which uses the Vercel deployment for this PR, I tested it successfully on https://github.com/gr2m/sandbox

PR #557, opened 2022-02-25
chore: move probot to devDependencies

hongaar:
This moves the `probot` dependency to `devDependencies` as it is only used in a type annotation. 

Rationale: I use `probot/settings` as a dependency in a project, and we get a lot of transient dependencies from the `probot` dependency which are not required to consume this code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,11970 @@
 {
   "name": "probot-settings",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "probot-settings",
+      "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


hongaar:
I will re-open this without changing the lock file version if there's support to merge this.

travi:
i think i have to say that supporting this type of use case is currently outside of the scope that we can commit to. while i can envision getting to a point where we would pull some of the base logic out to a library that could be consumed separately, we are not prepared to take a step like that yet. i wouldn't suggest relying on being able to depend on this repo as a library in its current state, but i also cannot currently offer a better solution if the app does not satisfy your needs.

hongaar:
FWIW I've ported parts of the code in this repository to TypeScript here: https://github.com/exivity/actions/tree/main/sync-defaults/src/settings

PR #571, opened 2022-04-18
Add prettier-standard auto-formatting

mattwynne:
## What's changed?

- Added prettier-standard and configured it for minimal impact
- Reformat files as part of `npm test`
- Run lint/formatting check in CI
- Run prettier-standard on all files to reformat them

## Feedback wanted

* Is this useful?
* Is there another library / technique I hadn't considered (prettier is my go-to tool for this, so I haven't looked much further)
* Should we run the `prettier-standard --check` in a separate CI job for clearer feedback in PRs?


mattwynne:
Based on the feedback in https://github.com/probot/settings/pull/569#pullrequestreview-943486336, I figured this might be a useful first step.

I have very weak opinions about the "right" formatting for code, but I do love using `prettier` to make it consistent. It seems that `standard` on it's own doesn't impose a line length, but if we use it with `prettier` we can ensure we get files in a consistent format every time.

.editorconfig
```diff
@@ -7,3 +7,4 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+max_line_length = 100
```
travi, line 10: i wasnt aware that editorconfig supported this property. i'm wondering what the behavior is when it "[forces hard line wrapping](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length)". i'm used to getting a lint message for violations of this and having the opportunity to figure out what would be sensible for a place to break the line. this sounds like it would likely break wherever it can. would it be better to leave this to prettier to figure out for us or resolve manually?

if we do go forward with this, i usually go with a length of 120. it tends to be a middle ground between folks that want longer lines and those that want to have side panels not result in side scrolling. does 100 have any specific significance with these tools?

.github/workflows/node-ci.yml
```diff
@@ -19,4 +19,5 @@ jobs:
           node-version: 16
           cache: npm
       - run: npm ci
+      - run: npx prettier-standard --check
```
travi, line 22: i'd like the checks that run in ci to align as much as possible with those run locally before pushing. could this be included in the list run by `npm test` (possibly as the `lint:js` script)?

package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "nodemon",
     "start": "probot run ./index.js",
     "test": "npm-run-all --print-label --parallel lint:* --parallel test:*",
-    "lint:js": "standard",
+    "lint:js": "prettier-standard --format --lint",
```
travi, line 11: i usually prefer an approach where the lint scripts only verify and do not automatically apply formatting. that way the fix/format step is a little more intentional and the lint script can run the same way with a manual run of `npm test` as it does on CI. would it make sense to move the auto-format piece to `lint:js:fix`?

travi, line 11: i assume this cli tool is available through some tool that we already depend on since there is no added dependency as part of this. i usually find it safer to declare tools that we use directly as direct (dev) dependencies. could this be added?

travi:
> I have very weak opinions about the "right" formatting for code, but I do love using `prettier` to make it consistent.

i have opinions, but am of the opinion that no ones opinions are more important than consistency. i very much agree with the philosophy of `prettier` with that goal, even though I don't have that much experience with `prettier` myself. i like moving in this direction.

i'm not at all married to using standard. that was inherited with this project. at some point, i would like to put together a more thorough eslint config (along with prettier) to further encourage consistency without a bunch of PR review discussions about individual preferences. i dont think going down that path needs to be part of this, but with that context, i think the steps you took are a good middle ground to take us a step closer.

travi:
> * Should we run the `prettier-standard --check` in a separate CI job for clearer feedback in PRs?

i'm interested to better understand what you mean by this. do you have examples of the type of feedback that could be surface with this approach?

.editorconfig
mattwynne: This file was a new discovery for me!

I added this just as a [place to configure prettier](https://prettier.io/docs/en/options.html#print-width). I think if we go for 120 that will be fine too. I was trying to find a middle ground to get prettier to make similar decisions to what's already in the code.

I think what you'll find in practice is that prettier will make a sensible decision about breaking the line well before 120 characters.

package.json
mattwynne: Sure, we can do that.

FWIW, I am used to having the auto-format [run on save in my editor](https://marketplace.visualstudio.com/items?itemName=numso.prettier-standard-vscode).

mattwynne: Resolved in b28583ff5b4e05f5a23de57cfcdc885a2b8cfd4f

mattwynne:
> > * Should we run the `prettier-standard --check` in a separate CI job for clearer feedback in PRs?
> 
> i'm interested to better understand what you mean by this. do you have examples of the type of feedback that could be surface with this approach?

Well, I'm just thinking that if you run the format/linting check and the tests as separate CI jobs (still in the same workflow) I think you'd see them as separate "checks" in the PR feedback. So if you forgot to reformat your code it might be easier to pick out that mistake rather than having to pick through the general "build" job's output. I'm just thinking about minimizing friction.

mattwynne:
> > > * Should we run the `prettier-standard --check` in a separate CI job for clearer feedback in PRs?
> > 
> > 
> > i'm interested to better understand what you mean by this. do you have examples of the type of feedback that could be surface with this approach?
> 
> Well, I'm just thinking that if you run the format/linting check and the tests as separate CI jobs (still in the same workflow) I think you'd see them as separate "checks" in the PR feedback. So if you forgot to reformat your code it might be easier to pick out that mistake rather than having to pick through the general "build" job's output. I'm just thinking about minimizing friction.

Since we've moved the reformatting to `lint:js:fix` I can pull this out of the CI workflow anyway.

package.json
mattwynne: I missed that!

.github/workflows/node-ci.yml
mattwynne: I notice that the `npm test` command appears to run all the `lint:*` commands, so presumably we'll both check *and* reformat... is this OK?

package.json
mattwynne: Fix in ac4f30f27688cad37a486eb75685e7534ffbf719

mattwynne:
@travi this is now reduced to:

- [adding max line length to `.editorconfig`](https://github.com/probot/settings/pull/571/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5b)
- [adding a `.prettierignore` so we only reformat JavaScript files](https://github.com/probot/settings/pull/571/files#diff-b640b344ee7f3f03d2a443795a5d0708ef50e2e6e34214109ab2aad13ad6ba98)
- [adding an `lint:js:fix` npm task](https://github.com/probot/settings/pull/571/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L11)
- [adding the dependency on `prettier-standard`](https://github.com/probot/settings/pull/571/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R38)
- a bunch of formatting changes

It's still pretty hard to read the diff with all the formatting changes so I figured I'd pick out the highlights!

.github/workflows/node-ci.yml
mattwynne: I tested this, and in fact it runs the `lint:js` task first, so if there are formatting problems it fails and stops there, and never runs the `list:js:fix` task.

So I think we get the behaviour you're looking for.

We maybe need to document somewhere that people can run `npm run lint:js:fix` to sort out formatting issues. Where would be a good place for that?

travi:
> It's still pretty hard to read the diff with all the formatting changes so I figured I'd pick out the highlights!

much appreciated

mattwynne:
In order to easily merge this into other branches, I propose I rebase this branch to have just two commits:

1) add all the config
2) apply it and reformat the files

That way we can easily cherry-pick the first commit into other live branches, then run the formatter, to reduce merge conflicts.

WDYT?

mattwynne:
OK that's done. I think the build should fail on the first commit, since it adds a CI check for formatting compliance, which isn't fixed until the second commit.

mattwynne:
> i'd also like to keep track of the commit that applies the reformatting, so that it can be ignored from blame in the future, similar to https://michaelheap.com/git-ignore-rev/. would you mind either including that in this PR or in a follow up?
> 
> > **[A better `git blame` with `--ignore-rev`](https://michaelheap.com/git-ignore-rev/)**Automatically ignore any automated changes in `git blame` by ignoring specific commits

Oooh OK will take a look. Didn't know about that - neat!

travi:
> I think the build should fail on the first commit, since it adds a CI check for formatting compliance, which isn't fixed until the second commit.

not ideal in cases like `git bisect`, but i do agree that it is appropriate in this case

mattwynne:
I appreciate your courage in making this change @travi - I think it will be worth it!

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+8e3d7f1a0114735005976434eb2b6baba9097bbd
```
travi, line 1: one last small request. lets include a comment that describes what the change was so that we know why that commit was ignored without needing to track it down in the actual history

travi: looks great!

Issue #566, opened 2022-03-25
Settings unchanged without errors


jftanner:
## Problem Description
I'm running a copy of `probot/settings` against a GitHub Enterprise instance. It used to work, but isn't now. Unfortunately, I don't know when it broke or why, because we don't update settings very often.

The bot appears to be getting events properly, but the settings aren't updated and no errors are logged. (I've got `LOG_LEVEL=trace` set in the container environment.)

Looking at the code, it seems to me like it _must_ be calling `syncSettings(context)`, but I have no insight past that. If necessary, I can fork the entire repo, add a bunch of logging, and deploy it into my cluster... but I'm hoping it won't come to that.

Any help would be greatly appreciated.

### What is actually happening

- The Github App sends an event payload.
- Probot receives the event and logs it: `INFO (http): POST / 200 - 527ms`
- If the settings file has changed: nothing else happens.

Note: As expected, I do get the following logs when settings shouldn't be updated:
- `DEBUG (probot): No changes in '.github/settings.yml' detected, returning...`
- `DEBUG (probot): Not working on the default branch, returning...`

### What is the expected behavior
- The Github App sends an event payload.
- Probot receives the event and logs it.
- Probot either logs an error, or updates the settings.

### Error output, if available

None. Which, unfortunately, is the problem.

### Are you using the hosted instance of probot/settings or running your own?
Running my own, with the following (mildly redacted) `package.json`:
```JSON
{
  "name": "[Redacted for security]",
  "version": "0.0.0-semantic-release-0",
  "private": true,
  "description": "Probot App that manages settings for Github repos",
  "contributors": [
    "James Tanner <[Redacted for privacy]>"
  ],
  "scripts": {
    "start": "probot run probot-settings",
    "build": "[Redacted for security]"
  },
  "dependencies": {
    "probot": "12.2.2",
    "probot-settings": "probot/settings"
  }
}
```
(Note: In the deployed container, `version` is actually `4.0.0`)

### If running your own instance, are you using it with github.com or GitHub Enterprise?
Github Enterprise

#### Version of probot/settings
Probot v12.2.2 (Node.js: v17.8.0)

#### Version of GitHub Enterprise
GHE 3.1.16

jftanner:
Here's a snippet of server logs, right after deploying a new instance:
<img width="417" alt="image" src="https://user-images.githubusercontent.com/10095474/160210761-12b1e5c5-c81c-416c-901b-b60df1151bc0.png">

[...]

PR #579, opened 2022-05-02
Add additional debugging output

jftanner:
This PR adds some debugging output for when `context.config` returns an empty object (or `null`). This resulted from #566, where I had neglected to set `GHE_HOST` so Octokit called public Github instead of GHE. Hopefully this debugging helps future users identify the same situation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,9 +6,18 @@ import SettingsApp from './lib/settings'
  */
 export default (robot, _, Settings = SettingsApp) => {
   async function syncSettings (context, repo = context.repo()) {
+    const { host } = context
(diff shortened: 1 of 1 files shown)


travi:
is there documentation that would be worth linking to in this log output that would have given hints toward why the config was unable to be read?

jftanner:
That's a good question. This is the relevant documentation: https://probot.github.io/docs/configuration/

Though, maybe I should add an additional check that verifies that the URLs in the webhook payload match `GHE_HOST` variable. That'd be a known failure state with a known fix, whereas there might be other reasons why `context.config` might fail.

Issue #613, opened 2022-07-17
enable renovate


travi:
i'd like to enable renovate over dependabot, now that we've moved the project to the new org. my default approach is pretty aggressive, auto-merging essentially everything and improving the test suites if dependency updates ever highlight a gap.

in my other projects, i use one of the options that i've defined in https://github.com/form8ion/renovate-config as the config to extend from. @gr2m, i'm curious if you think it makes sense to extend one of those here, or if we should define separate configuration for this org.

as far as dependabot goes, i think it makes sense to keep it active until we get renovate configured to a level that we are happy with, but we might as well scale dependabot back to at least vulnerability patches only after that.

gr2m:
Happy to try new things 👍🏼 I like dependabot for apps, but happy to use renovate. `js-app.json` looks good to me: https://github.com/form8ion/renovate-config/blob/master/js-app.json

travi:
> what’s wrong with Dependabot?

for me, it isnt that there is anything wrong with dependabot. it is more that renovate does more to help remove some burdens from our maintainers. it is more configurable if we want to treat different categories of dependencies differently. it can be configured to auto-merge. it can update the .nvmrc for the preferred node version for the project.

plus, i've simply used renovate more heavily and find that it gets in my way less than dependabot, again improving the maintenance burden

localheinz:
@travi 

For the record, I use a dedicated workflow for automatically merging pull requests created by Dependabot:

https://github.com/ergebnis/php-package-template/blob/d729553587c84a2495b602fbc0ebeb3d19764fc7/.github/workflows/merge.yaml

travi:
@localheinz thanks for the reference. 

i have gone at least partially down similar paths, including trying a few tools like [fastify/github-action-merge-dependabot](https://github.com/fastify/github-action-merge-dependabot). 

in the end, renovate still wins in doing the most for me without adding to my maintenance burden. i can configure it to automerge easily and can even [configure it to use GitHub's native automerge capability](https://docs.renovatebot.com/key-concepts/automerge/#faster-merges-with-platform-native-automerge). I dont need to maintain a separate bot user account and personal access tokens for that account. i can even install [renovate-approve](https://github.com/apps/renovate-approve) to handle the automated approvals where needed (or even [renovate-approve-2](https://github.com/apps/renovate-approve-2) if multiple automated reviews are needed)

PR #615, opened 2022-07-18
ci(renovate): extended the application config from form8ion

travi:
for #613


travi:
this defines the renovate config. once this is in place, we can install the app(s)

gr2m (approved):

PR #623, opened 2022-07-28
ci(release): defined a release workflow

### LINKED ISSUE #614: releases
#613 got me thinking a little bit about how what we consider this project to be. since probot apps can both be apps but also dependencies for other probot apps i'm a little conflicted on which way to treat this app as far as dependency pinning goes. i typically follow the renovate recommendation to pin all dependencies for apps, but leave definitions as ranges for libraries. 

we dont currently publish versioned releases, but we definitely have users running their own instances. @gr2m, you do a good job of handling releases for your probot apps, like https://github.com/wip/app/releases, so i'm curious about your thoughts on this. also, it feels a little odd to start versioning at v1.0.0 now since the app has been around for a while, but i don't have a better suggestion than that either 🤷🏼

travi:
based on https://github.com/wip/app/blob/master/.github/workflows/release.yml

closes #614

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,19 @@
+name: Release
+"on":
+  push:
+    branches:
+      - master
+jobs:
+  release:
+    name: release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+      - run: npm clean-install
+      - run: npx semantic-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -1,0 +1,7 @@
+{
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
+    ["@semantic-release/github", {"successComment": false}]
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "probot-settings",
-  "version": "1.0.0",
+  "version": "0.0.0-semantically-released",
+  "private": true,
   "description": "",
   "repository": "github:repository-settings/app",
   "main": "index.js",


travi:
a few thoughts to consider before merging. i'm interested in any lean that might sway in any particular direction.

* merging without any further action will create the first release as v1.0.0 and consider every commit before now in the changes listed in that release (and comment on all of them)
* the app has existed for some time with plenty of previous changes along the way, so v1.0.0 now seems a bit odd, especially since little about this particular timing suggests that it has reached any specific level of maturity to be promoted from pre-1.0.0 to "stable"
* we could add a tag for an arbitrary version that is something different than v1.0.0, but some other arbitrary version doesnt have any particular significance either. it would avoid having comments added to all of the previous PRs though
* https://github.com/repository-settings/app/pull/616 will raise the minimum required node version defined in `engines.node`, which will be a breaking change. does it make more sense to release v1.0.0 and _then_ merge that PR, resulting in v2.0.0 shortly after v1.0.0, or is that the "significant" change that results in v1.0.0 happening. if the latter, the breaking detail would get lost in the generated release notes, but those could be edited down to make the breakage be more clear

gr2m:
Maybe set `successComment: false` in the configuration? That seems to disable success comments?
https://github.com/semantic-release/github/blob/f4fea60b853879210d9c9dab02a9db14d72a5b16/lib/success.js#L41

Regarding https://github.com/repository-settings/app/pull/616, I think we should release v1 and then shortly after v2. It will make the breaking change transparent. I think by far the most users won't care and the ones that do will appreciate it

travi:
> Maybe set `successComment: false` in the configuration?

yeah, i was thinking this would be the best bet if we decided to start at v1.0.0, but didnt want that suggestion to influence any reason to consider starting at a different version. i'm in line with this and will update the config later today.

should we remove the custom config after the initial release then? i think the comments can be valuable under normal circumstances

> Regarding #616, I think we should release v1 and then shortly after v2. It will make the breaking change transparent. I think by far the most users won't care and the ones that do will appreciate it

i think this is where i was leaning too, but this helps solidify those thoughts.

travi:
i got the config added that i'm mostly sure will do the trick. please do take a close look at it though, since i don't use custom config very often for semantic-release and could have easily overlooked something.

gr2m (approved):
Looks good to me 👍🏼

PR #671, opened 2022-11-22
feat: Add support for environments

### LINKED ISSUE #607: Support for Github Environments
## Prerequisites:

* Is the functionality available in the GitHub UI? If so, please provide a link to information about the feature.
https://github.com/{org}/{repo}/settings/environments

* Is the functionality available through the GitHub API? If the functionality is available, please provide links to the
  API documentation (https://developer.github.com/v3/) as well as the Octokit documentation (https://octokit.github.io/).
-  https://docs.github.com/en/rest/deployments/environments

* If the functionality is not yet available in the API, it would be helpful if you
  contacted support (https://support.github.com/) or posted in the Community Forum (https://github.community/). Please
  include a link to the forum post if you create one or a copy of the response from support.

## New Feature

Please describe the desired new functionality:
Add support to manage Github environments with settings.yml

rlmartin:
Fixes #607

I tested this manually (using a private GitHub App + [test repo](https://github.com/rlmartin/test-repository-settings-app) + smee.io) and it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ collaborators:
   # * `triage` - Recommended for contributors who need to proactively manage issues and pull requests without write access.
 
 # See https://docs.github.com/en/rest/deployments/environments#create-or-update-an-environment for available options
-# Note: deployment_branch_policy differs from the API for ease of use. Either protected_branches (boolean) OR custom_branches (array of strings) is required; this will manage the API requirements under the hood. See https://docs.github.com/en/rest/deployments/branch-policies for documentation of custom_branches. If both are provided in an unexpected manner, protected_branches will be used.
+# Note: deployment_branch_policy differs from the API for ease of use. Either protected_branches (boolean) OR custom_branches (array of strings) can be provided; this will manage the API requirements under the hood. See https://docs.github.com/en/rest/deployments/branch-policies for documentation of custom_branches. If both are provided in an unexpected manner, protected_branches will be used.
 # Either removing or simply not setting deployment_branch_policy will restore the default 'All branches' setting.
 environments:
   - name: production
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ collaborators:
   # * `triage` - Recommended for contributors who need to proactively manage issues and pull requests without write access.
 
 # See https://docs.github.com/en/rest/deployments/environments#create-or-update-an-environment for available options
+# Note: deployment_branch_policy differs from the API for ease of use. Either protected_branches (boolean) OR custom_branches (array of strings) is required; this will manage the API requirements under the hood. See https://docs.github.com/en/rest/deployments/branch-policies for documentation of custom_branches. If both are provided in an unexpected manner, protected_branches will be used.
 environments:
   - name: production
     wait_timer: 5
@@ -127,11 +128,11 @@ environments:
         type: 'User'
     deployment_branch_policy:
       protected_branches: true
-      custom_branch_policies: false
   - name: development
     deployment_branch_policy:
-      protected_branches: false
-      custom_branch_policies: true
+      custom_branches:
+        - main
+        - dev/*
 
 # See https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions for available options
 teams:
--- a/README.md
+++ b/README.md
@@ -116,6 +116,23 @@ collaborators:
   # * `maintain` - Recommended for project managers who need to manage the repository without access to sensitive or destructive actions.
   # * `triage` - Recommended for contributors who need to proactively manage issues and pull requests without write access.
 
+# See https://docs.github.com/en/rest/deployments/environments#create-or-update-an-environment for available options
+environments:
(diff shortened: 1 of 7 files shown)


travi:
you've done the steps that you need to. i just need to find time to review. this looks like a great contribution, so thank you for the effort you've put in. i'm hopeful that i can take a look by friday afternoon.

rlmartin:
👍 thanks so much

travi:
i did not get a chance to review this today, but still have it high on my list, so i hope to get to it soon

rlmartin:
Hi @travi - do you think you'll get a chance to review this soon? Thanks!

rlmartin:
FYI I just added a second commit because I realized that setting `deployment_branch_policy.custom_branch_policies = true` needs to be combined with setting a list of branches using the [deployment branch policies API](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28). I updated the tests (which are passing) and manually tested again (using smee.io).

rlmartin:
FYI after this is merged, the GitHub App settings will need to be updated to request `Read and write` access on `environments` in Repository permissions.

rlmartin:
Hi @travi - just checking in on a review. Thanks!

rlmartin:
Hi @travi I would really appreciate a review here, before this PR gets too stale and I risk possibly needing to make larger fixes to accommodate any change that has happened in the interim. Thanks!

travi:
sorry for the delay getting back to this. we've been working to get some releases out the door for semantic-release, so that work has taken my focus recently.

> the GitHub App settings will need to be updated to request `Read and write` access on `environments` in Repository permissions.

do you have a source for this detail? from https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment, it looks like `administration:write` is needed (which settings already asks for), but i don't see a requirement for additional permissions beyond that. Also, I was able to test locally and have it work an app definition that does not have any environments specific permissions. Maybe this has changed?

README.md
```diff
@@ -116,6 +116,24 @@ collaborators:
   # * `maintain` - Recommended for project managers who need to manage the repository without access to sensitive or destructive actions.
   # * `triage` - Recommended for contributors who need to proactively manage issues and pull requests without write access.
 
+# See https://docs.github.com/en/rest/deployments/environments#create-or-update-an-environment for available options
+# Note: deployment_branch_policy differs from the API for ease of use. Either protected_branches (boolean) OR custom_branches (array of strings) is required; this will manage the API requirements under the hood. See https://docs.github.com/en/rest/deployments/branch-policies for documentation of custom_branches. If both are provided in an unexpected manner, protected_branches will be used.
```
travi, line 120: https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment shows an option to allow all branches by setting the `deployment_branch_policy` to `null`. is that enabled with this approach?

rlmartin:
Oh, hmm, I just kind of assumed it was required, since it is there in the list of permissions when setting up the app (which I did to test this). If it works with _only_ `administration:write`, great - but I didn't see `environments` listed there at all.
Screenshot 2023-01-27 at 6 16 18 PM

README.md
travi: @rlmartin not sure if you saw this question. i think the rest of the PR looks good, but i think we should handle this if it isnt already.

rlmartin: Good question. I'll take a look and get back to you.

rlmartin: Good catch - subtle difference between `undefined` and `null` in the GitHub API. I fixed it now and manually verified that it works (and updated the documentation here).

rlmartin:
@travi I removed the `environments:write` permission from my test app and it still worked. So I agree that there is no need to update the App's requested settings - `administration:write` appears to be sufficient. Not sure why there is that extra `environments:write` there 🤷

rlmartin:
@travi I believe I have addressed both of your comments and this is ready for re-review.

README.md
```diff
@@ -116,6 +116,25 @@ collaborators:
...
```
travi, line 120: > (and updated the documentation here)

i think we should make this small adjustment now that null is allowed

```suggestion
# Note: deployment_branch_policy differs from the API for ease of use. Either protected_branches (boolean) OR custom_branches (array of strings) can be provided; this will manage the API requirements under the hood. See https://docs.github.com/en/rest/deployments/branch-policies for documentation of custom_branches. If both are provided in an unexpected manner, protected_branches will be used.
```

travi (changes requested):
one more small tweak to the docs seems worthwhile. i think we are ready to merge with that change

README.md
rlmartin: Sounds good. Yes setting `deployment_branch_policy: null` will result in `All branches`.

travi (approved):
thanks a lot for this contribution, @rlmartin!

rlmartin:
@travi thank you!

PR #747, opened 2023-06-16
test(integration): extracted common logic for setting up the proper settings content for each scenario

travi:
for #145

### DIFF
--- a/test/integration/common.js
+++ b/test/integration/common.js
@@ -1,6 +1,11 @@
+const path = require('node:path')
+const { promises: fs } = require('node:fs')
+const { OK } = require('http-status-codes')
 const { Probot } = require('probot')
+
 const nock = require('nock')
 const any = require('@travi/any')
+
 const settingsBot = require('../../index')
 const settings = require('../../lib/settings')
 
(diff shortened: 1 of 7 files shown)


travi:
hmm, build failure appears to be related to https://github.com/thlorenz/find-parent-dir/issues/8, but it is unclear to me why it shows up specifically for this PR without being an existing problem for previous builds 🤔

travi:
it appears that we are pulling in that dependency through `prettier-standard`:

```
$ npm ls find-parent-dir
probot-settings@0.0.0-semantically-released /path/to/repository-settings/app
└─┬ prettier-standard@16.4.1
  └─┬ prettierx@0.11.3
    └── find-parent-dir@0.3.0
```

we are using the latest version of `prettier-standard`:

```
$ npm dist-tags ls prettier-standard
beta: 14.0.0-beta.5
latest: 16.4.1
```

the dependency on `prettierx` from `prettier-standard` is an exact dependency, which in turn defines an exact dependency on `find-parent-dir`, preventing updating to the fixed version of `find-parent-dir` by just updating our lockfile.

travi:
> hmm, build failure appears to be related to [thlorenz/find-parent-dir#8](https://github.com/thlorenz/find-parent-dir/issues/8), but it is unclear to me why it shows up specifically for this PR without being an existing problem for previous builds 🤔

in the process of defining an override for the version of `find-parent-dir`, i realized that i was overlooking legit lint errors. the issue related to `find-parent-dir` is just a warning and does not fail the build. i still think we should move away from it, but that explains why it wasnt causing actual problems for us us to this point.

gr2m (approved):

gr2m:
> `prettier-standard` was last published over 3 years ago, so i'd suggest we move away from it. i'd prefer to avoid making that change as part of this PR, though.

💯

PR #795, opened 2023-09-12
Adding ability to manage tag protections via settings

### LINKED ISSUE #772: Add support for adding Protected Tags
## Prerequisites:

* Is the functionality available in the GitHub UI? If so, please provide a link to information about the feature.
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules

* Is the functionality available through the GitHub API? If the functionality is available, please provide links to the
  API documentation (https://developer.github.com/v3/) as well as the Octokit documentation (https://octokit.github.io/).
  https://docs.github.com/en/free-pro-team@latest/rest/repos/tags?apiVersion=2022-11-28#create-a-tag-protection-state-for-a-repository

## New Feature

Adding the ability to add protected tags via the .github/settings.yml file. Similar to how branches and branch protection is added,  adding the ability to specify a list of patterns to add as protected tags.
e.g.
```
tags:
  - protection:
       pattern: ["v1", "release/*, "latest"] 
```

ctoestreich:
This should allow tag protection

* Fixes #772

### DIFF
--- a/test/fixtures/tags-config.yml
+++ b/test/fixtures/tags-config.yml
@@ -1,0 +1,3 @@
+tag_protection:
+  - pattern: "v1"
+  - pattern: "duplicate"
--- a/README.md
+++ b/README.md
@@ -148,6 +148,11 @@ teams:
   - name: docs
     permission: push
 
+# See https://docs.github.com/en/rest/repos/tags#create-a-tag-protection-state-for-a-repository for available options
+tags:
+  - pattern: "v1"
+  - pattern: "release/*"
+
 branches:
   - name: master
     # https://docs.github.com/en/rest/reference/repos#update-branch-protection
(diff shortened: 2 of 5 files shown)


test/fixtures/tags-config.yml
```diff
@@ -0,0 +1,3 @@
+tag_protection:
```
travi, line 1: lets simplify this to `tags` to be more similar to the existing `branches`, for example

test/fixtures/tags-config.yml
```diff
@@ -0,0 +1,3 @@
...
+  - pattern: "v1"
```
travi, line 2: i don think including `pattern` here is helpful since it is the only property. lets simplify to just be the pattern

travi (changes requested):
a couple of tweaks to the shape of the proposed config section. also, could you please also update the [configuration docs](https://github.com/repository-settings/app/blob/master/docs/configuration.md) to include an example of this functionality?

travi:
also worth considering: https://github.com/repository-settings/app/issues/772#issuecomment-1986455617

PR #847, opened 2024-01-12
feat(package): publish the app as an npm package

travi:


docs/configuration.md
travi, line 1: i plan to give more attention to improving the details in this doc, but wanted to get it moved to its own file so that the readme was less noisy and folks can find their way to getting started more easily

docs/self-host.md
```diff
@@ -0,0 +1,88 @@
...
+    await server.stop();
+  });
+
+  await server.load(app);
+
+  await server.start();
+};
+
+start();
+```
```
travi, line 60: interested in whether you think there is a simpler example that would help folks understand the goal, or even a more generic probot doc to link to, but wanted to provide some level of guidance here

docs/self-host.md
```diff
@@ -0,0 +1,88 @@
...
+
+This option offers you the most flexibility
+
+#### Depend on the package from npm
+
+```shell
+npm install @repository-settings/app
+```
+
+#### Example node.js app
```
travi, line 23: in addition to this sort of example, i could see value in providing a minimal example for deploying to a serverless environment. i dont have as much experience with that approach lately, so open to suggestions. i dont think adding that needs to block getting this progress merged, but could be a valuable follow up

docs/self-host.md
```diff
@@ -0,0 +1,88 @@
...
+- Actions: **Read only**
+
+### Organization Permissions
+
+- Members: **Read & Write**
+
+### Events
+
+- Push
+- Repository
```
travi, line 86: i don't know how accurate these even still are, but wanted to get them moved into this doc as-is for now. can adjust for correctness as a follow up. i also realized that we also havent added the file to enable the manifest flow, which is probably the better source of truth to maintain, so maybe that is the better follow up here

docs/deploy.md
```diff
@@ -1,25 +0,0 @@
-# Deploying
-
-If you would like to run your own instance of this plugin, see the [docs for deploying plugins](https://github.com/probot/probot/blob/master/docs/deployment.md).
```
travi, line 3: i dont think the plugins approach is even valid anymore, so i captured the idea of this more as just forking the repo.

README.md
```diff
@@ -1,216 +1,66 @@
...
 
-This GitHub App syncs repository settings defined in `.github/settings.yml` to GitHub, enabling Pull Requests for repository settings.
+## Table of Contents
+
+* [Usage](#usage)
+  * [Install](#install)
+    * [Hosted GitHub.com App](#hosted-githubcom-app)
+    * [Self-Hosted App](#self-hosted-app)
+  * [Configuration](#configuration)
+* [Security Implications](#security-implications)
```
gr2m, line 19: I usually leave out a TOC these days, because GitHub has its own build into the UI now, and unless updating a TOC in a markdown file is automated it tends to get out of sync with the actually sections.

docs/self-host.md
gr2m: I think it's a good idea, but we can do a follow up for that.

docs/self-host.md
gr2m: I think if you don't need any custom functionality and you want a long-running process, I think there is no code necessary at all, you should be able to do `"start": "probot run node_modules/@repository-settings/app/index.js"`, or maybe `@repository-settings/app` should expos its own binary like `"start": "repository-settings"`?

If users need more customization, for example for a serverless environment, then we can point them to https://probot.github.io/docs/development/#run-probot-programmatically and maybe provide an example for a serverless app?

gr2m (approved):
I couldn't try out the code but from looking at it this is great, I'd move forward and then iterate on it as needed

README.md
travi: thats fair. my reason for including is usually more for drawing attention to sections below the fold. i've known about the built-toc, but it is so hidden that i actually forgot it is even there, so i dont know that it accomplishes that goal. 

updating this one is fairly automatic because of the `generate:md` script that uses remark to update. i dont have a way to fully automate in mind, but i've wanted to investigate a way to fail verification if running the generate script would result in a diff.

open to more feedback on this, but lets address after getting his iteration merged

docs/self-host.md
travi: i like the idea of including a few different variations from simplest to most customizability. 

> maybe `@repository-settings/app` should expos its own binary like `"start": "repository-settings"`?

i like this idea

travi:
merging this iteration, with the intention to follow up on the remaining comments in later iterations

PR #854, opened 2024-01-26
test(integration): convert from jest + nock to cucumber + msw

travi:
since nock isnt currently able to stub native fetch calls, using it in the integration tests prevents use of the latest octokit/probot versions. the primary goal of this effort is to migrate the github api call stubs from nock to msw to overcome that limitation and enable upgrading to probot v13.

while i'm going through the effort, i'm also converting the jest _integration_ tests to use cucumber instead. i find that cucumber is more natural for large scope tests like these, both making the macro-behaviors more understandable and organizing repetitive setup logic. i think there is opportunity to further clean up some of the repetitive setup logic beyond what i've done as part of this initial effort, but i'm still somewhat new to msw and i'm focusing more on making the tests functional as this first step and can circle back later to further refactor.

i'm not touching the unit tests for now since i think jest is fine for unit/micro tests. when we get to migrating to esm, i think we will likely want to migrate those to vitest simply because of improved esm support. jest and vitest are api compatible, so that will be a smaller effort and keep essentially the same dev experience as current.


travi:
current failure when testing against node v16:

```
/home/runner/work/app/app/node_modules/msw/src/core/HttpResponse.ts:37
export class HttpResponse extends Response {
                                   ^

 ReferenceError: Response is not defined
     at null.<anonymous> (/home/runner/work/app/app/node_modules/msw/src/core/HttpResponse.ts:37:35)
     at ModuleJob.run (node:internal/modules/esm/module_job:175:25)
     at async Loader.import (node:internal/modules/esm/loader:178:24)
     at async importModuleDynamicallyWrapper (node:internal/vm/module:437:15)
     at async importer (/home/runner/work/app/app/node_modules/@cucumber/cucumber/lib/importer.js:10:10)
     at null.getSupportCodeLibrary (/home/runner/work/app/app/node_modules/@cucumber/cucumber/src/api/support.ts:33:5)
     at null.runCucumber (/home/runner/work/app/app/node_modules/@cucumber/cucumber/src/api/run_cucumber.ts:45:9)
     at Cli.run (/home/runner/work/app/app/node_modules/@cucumber/cucumber/src/cli/index.ts:79:25)
     at Object.run (/home/runner/work/app/app/node_modules/@cucumber/cucumber/src/cli/run.ts:33:14)
```

`Response` came into node with the fetch api in v18. we're overdue to drop node v16 support in this app, so i'm ok doing that as part of this effort, but would have preferred merging this independently of that.

if we will need to release a major version because of that breaking change, i'd prefer to also update to probot v13 as well. that will make this PR even bigger, but probably the right move

gr2m:
the biggest pain of upgrading to Probot v13 is incompatibility with nock due to the native fetch. That should not be a problem with the move to msw. So I hope it will be pretty smooth

gr2m:
> since nock isnt currently able to stub native fetch calls

try `npm install --save-dev nock@beta` ✨

gr2m (approved):
love the eloquence of cucumber tests!

travi:
> love the eloquence of cucumber tests!

i find them so much easier to define clear scenarios for larger scoped tests. i think this will make it much easier to reason about the behaviors within the app

PR #855, opened 2024-01-27
next major version

travi:
- [x] drop support for node v16
- [x] upgrade to probot v13

BREAKING CHANGES: 
* support for node versions v16, v17, and v19 has been dropped
* BREAKING CHANGE: probot v13 includes breaking changes. see the probot release notes for details: https://github.com/probot/probot/releases/tag/v13.0.0


travi:
we will need to coordinate handling the breaking webhook path change and the vercel pre-parsing for the hosted instance before merging this, but i think the changes are otherwise ready to be promoted

gr2m (commented):
Changes look good to me 👍🏼 

> the vercel pre-parsing for the hosted instance before merging this

yeah I had problems with that as well, I couldn't figure out how to programmatically disable it, only using the environment variable. But I also didn't spend much time on investigating. Definitely do a pre-deployment and test it first before

travi:
> I couldn't figure out how to programmatically disable it, only using the environment variable

since we publish this app as a package, i think it makes sense to limit the directive to disable to the vercel deployment, so the [environment variable approach](https://vercel.com/docs/functions/runtimes/node-js#disabling-helpers-for-node.js) makes sense to me for now. 

i could see value in having a separate repo at some point in the future that is dedicated to installing the package and deploying to vercel. that would enable the programmatic approach making more sense and also enabling us to release a package version separately from deploying in case there are ever vercel specific issues that we need to handle separately. hopefully that is rare, but i like the ability to handle that and also eat our own dog food by consuming the package.

> Definitely do a pre-deployment and test it first before

i'll test out the preview for this beta branch

travi:
> i'll test out the preview for this beta branch

i got a test app instance and used it for the preview deployed to the beta branch. without the secrets defined for that instance, the ping webhook was failing the signature check. after defining and re-deploying, that webhook was processing correctly. 

> the breaking webhook path change

looks like our config was already using this path, so i see no change necessary. i assume that was able to be configured that way before because of it being a serverless deployment?

> vercel pre-parsing

i configured the `NODEJS_HELPERS` variable for both the preview instance and the production instance. i've seen no errors in the logs for either instance after getting the signature piece sorted out. i think it is safe to leave it defined for production, but i havent tested a webhook against the the preview instance other than the ping yet (which wouldnt actually exercise the app) and need to step away. if we think we need to exercise that, we'll need to set up a quick test repo and make sure the preview instance is installed there

travi:
i was able to get a test repo in place and wired up to the preview installation, but it is resulting in 401 responses for actual changes, so working on tracking down what is resulting in that

travi:
> looks like our config was already using this path, so i see no change necessary. i assume that was able to be configured that way before because of it being a serverless deployment?

forgot the details of the vercel config that we are using based on https://probot.github.io/docs/deployment/#vercel, so that should be good to go

travi:
> it is resulting in 401 responses for actual changes, so working on tracking down what is resulting in that

i got this resolved and everything appears to be working as expected. i think this is ready to merge

gr2m (approved):

PR #910, opened 2024-04-04
feat(environments): custom branch policies support type ('branch', 'tag')

goruha:
## What
* Added `type` support for environment `deployment_branch_policy`  custom branches.

## Why
* Environment `deployment_branch_policy` supports custom branches type - `branch` or `tag`. The type is an option parameter that sets `branch` by default. These changes allow us to specify deployment_branch_policy for  `tag` 

## Config example

```yaml
environments:
  - name: development
    deployment_branch_policy:
      custom_branches:
        - dev/*
        - name: release/*
          type: branch
        - name: v*
          type: tag
```

You can specify `custom_branches` list item as `string` for back-compatibility or as object

```yaml
name: `string`
type: `branch | tag`
```

## Releated links
* [Create a deployment branch policy API](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28#create-a-deployment-branch-policy)

### DIFF
--- a/lib/plugins/environments.js
+++ b/lib/plugins/environments.js
@@ -38,7 +38,9 @@ function deploymentBranchPolicyToString (attrs) {
     return JSON.stringify(
       shouldUseProtectedBranches(attrs.protected_branches, attrs.custom_branches)
         ? { protected_branches: true }
-        : { custom_branches: attrs.custom_branches.sort() }
+        : {
+            custom_branches: attrs.custom_branches.sort((a, b) => JSON.stringify(a).localeCompare(JSON.stringify(b)))
+          }
     )
   }
 }
@@ -66,6 +68,14 @@ module.exports = class Environments extends Diffable {
       // Force all names to lowercase to avoid comparison issues.
       this.entries.forEach(environment => {
         environment.name = environment.name.toLowerCase()
+        if (environment.deployment_branch_policy && environment.deployment_branch_policy.custom_branches) {
+          environment.deployment_branch_policy.custom_branches = environment.deployment_branch_policy.custom_branches.map(
+            rule => ({
+              name: rule.name || rule,
+              type: rule.type || 'branch'
+            })
+          )
+        }
       })
     }
   }
@@ -88,7 +98,10 @@ module.exports = class Environments extends Diffable {
               environment.name
             )
             environment.deployment_branch_policy = {
-              custom_branches: branchPolicies.map(_ => _.name)
+              custom_branches: branchPolicies.map(_ => ({
+                name: _.name,
+                type: _.type
+              }))
             }
           } else {
             environment.deployment_branch_policy = {
@@ -152,8 +165,8 @@ module.exports = class Environments extends Diffable {
             org: this.repo.owner,
             repo: this.repo.repo,
             environment_name: attrs.name,
-            name: rule.name || rule,
-            type: rule.type || 'branch'
+            name: rule.name,
+            type: rule.type
           })
         )
       )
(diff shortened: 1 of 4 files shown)


goruha:
@julekirk @mayliang021 @travi 
may I have a review pls?

travi:
supporting tag targets feels separate from what makes sense under "custom_branches". would a separate section for tags be a better fit?

goruha:
@travi can we have this PR merged?

test/integration/features/step_definitions/environments-steps.mjs
```diff
@@ -75,7 +75,11 @@ Given('an environment exists with a {string} branches deployment branch policy',
   )
 
   if (policyType === 'custom') {
-    this.customBranches = any.listOf(() => ({ name: any.word(), id: any.integer() }))
+    this.customBranches = any.listOf(() => ({
+      name: any.word(),
+      id: any.integer(),
+      type: any.fromList(['branch', 'tag'])
```
travi, line 81: i appreciate that you tracked down how to accomplish this and i think this is totally valid for the scenario that defines `type`, but i think this change now results in not having a scenario tested for the option to not include type, doesnt it. it seems like we probably should define an additional scenario for including type in https://github.com/repository-settings/app/blob/master/test/integration/features/environments.feature

i also wonder if the inclusion of a type that doesnt match the default requires any additional care to avoid unnecessary update requests when the intended value matches what is in the api. could you verify that comparing an existing rule in the api avoids extra calls when the type is and is not present in the settings.yml? if that requires additional logic to avoid extra calls, we may need test scenarios for those cases as well

lib/plugins/environments.js
```diff
@@ -147,12 +147,13 @@ module.exports = class Environments extends Diffable {
...
       await Promise.all(
-        attrs.deployment_branch_policy.custom_branches.map(name =>
+        attrs.deployment_branch_policy.custom_branches.map(rule =>
           this.github.request('POST /repos/:org/:repo/environments/:environment_name/deployment-branch-policies', {
             org: this.repo.owner,
             repo: this.repo.repo,
             environment_name: attrs.name,
-            name
+            name: rule.name || rule,
+            type: rule.type || 'branch'
```
travi, line 156: because of the default values, i think we also need to handle similar in https://github.com/repository-settings/app/blob/36173c5eadad3084dc6b3153b466d834e167a7c5/lib/plugins/environments.js#L54-L59 to avoid repetitively updating rules that are already in an ok state.

travi:
> Yeah, the separate settings for tags look better.
> However, GitHub API does not have this separation, and we should follow the GitHub API structure.
> Mapping the GitHub API contract to `settings.yml `solves the lack of documentation - you can always reference GitHub API to see what fields are available.

in general, i agree. however, as mentioned in the [docs for the environments plugin](https://github.com/repository-settings/app/blob/master/docs/plugins/environments.md), we already deviate a bit, so that is the reason that gave me pause here. since even the github api still uses the branches terminology for this, i think i'm ok with proceeding with the proposed approach

travi (changes requested):

test/integration/features/step_definitions/environments-steps.mjs
goruha: Actually, we always specify `type` for API calls.
https://github.com/repository-settings/app/pull/910/files#diff-d3173e671826d864f026b433955eeef4008090a5fc6c3111f8ec2e3977e2bc4dR156

So, the test is correct and does not dismiss the case without type.

lib/plugins/environments.js
goruha: @travi, thanks for the nice catch.
I'll address this comment this week.

goruha: @travi, you were right; there was a bug related to repetitively updating.
I made the fix and added unit tests to ensure the bug was solved.
May I have a review?

P.S.: Sorry for the delayed response.

travi (approved):
thanks!

Issue #732, opened 2023-04-18
feature: add repositories rules/rulesets


electriquo:
## Prerequisites:

Following [Introducing Repository Rules Public Beta](https://github.blog/changelog/2023-04-17-introducing-repository-rules-public-beta/), GitHub also introduced a [Rules REST API endpoint](https://docs.github.com/en/rest/repos/rules), but it seems to still be missing from Octokit (please do verify).

## New Feature

Settings app to support the repository rules endpoint

travi:
initial support has been released in https://github.com/repository-settings/app/releases/tag/v4.1.0. [please note](https://github.com/repository-settings/app/blob/master/docs/plugins/rulesets.md#repository-rulesets) that this is considered a pre-release and may have breaking changes applied before it is considered fully released. feedback is welcome if you try it out in the meantime.

since this is not yet fully released, i'm going to keep this issue open and will follow up with more detail about the details i'm still considering further iteration on

loozhengyuan:
@travi Thanks for adding support for this! I have tried this and thus far it is working great for ruleset creations but faced issues when updating it.

I noted that adding a new rule to a ruleset does not update the rule. Initially, I thought that it might be an issue the request schema but the desired changes were able to be applied if I were to change the ruleset name. This meant that the schema is working fine but probably faced issues when updating it.

loozhengyuan:
I also noticed that the `strict_required_status_checks_policy` value must be defined, which seems to tally with the [API docs](https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#create-a-repository-ruleset). I am not sure if this should be solved in code or via documentation, but the example in the documentation does not work for me without making the following change:

```diff
rulesets:
  - name: verification must pass
    target: branch
    enforcement: active
    conditions:
      ref_name:
        include:
          - "~DEFAULT_BRANCH"
    rules:
      - type: required_status_checks
        parameters:
+         strict_required_status_checks_policy: false
          required_status_checks:
            - context: test
              integration_id: 123456
```

travi:
thanks for testing it out, @loozhengyuan!

> I noted that adding a new rule to a ruleset does not update the rule

i've noticed that i missed some details for the update scenario as well. i think i've identified the primary issue and have a partial solution that i'm working through already, so i think the results will be better once i get that finished up. unfortunately, i can't promise a timeline right now because i'm preparing to travel for work this week and it is questionable how much time i will find before returning home. this is high on my list once i have availability, though.

[...]

PR #1029, opened 2024-09-15
feat(rulesets): initial support repository rulesets (semi-dark release with the potential for breaking changes)

travi:
while still WIP for ergonomics, this is production ready, but with the potential for breaking changes before considered ready for full release. my plan is to release it semi-dark so that it can be used, but without highlighting it beyond the release notes or the specific docs page (which warns about the potential for breakages)

for #732

### DIFF
--- a/docs/plugins/rulesets.md
+++ b/docs/plugins/rulesets.md
@@ -1,0 +1,53 @@
+# Repository Rulesets
+
+See https://docs.github.com/en/rest/repos/rules#update-a-repository-ruleset for
+all available ruleset properties and a description of each.
+
+```yaml
+rulesets:
+  - name: prevent destruction of the default branch
+    target: branch
+    enforcement: active
+    conditions:
+      ref_name:
+        include:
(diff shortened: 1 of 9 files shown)


docs/plugins/rulesets.md
```diff
@@ -0,0 +1,59 @@
+# Repository Rulesets
+
+> [!WARNING]
+> Support for Repository Rulesets is still under development.
+> Details may still change, like how the configuration is defined in the `settings.yml`.
+> Please follow [#732](https://github.com/repository-settings/app/issues/732) for progress updates.
+> Feedback is appreciated, but adopt cautiously, with the expectation of breaking changes until support is fully released.
+
```
travi, line 8: while this should be fully functional with the current schema, it mirrors the github API expectations, which are somewhat cryptic for a non-programatic context. we prefer to mirror the github API most of the time, but i'm leaning away from that in this case

because of that, i want to release this progress in a semi-dark pre-release fashion. this warning is included in the docs to attempt to make that visible

PR #1043, opened 2024-09-29
docs(rulesets): fix ruleset status checks example

loozhengyuan:
This commit fixes the example in the ruleset documentation, which was missing the `strict_required_status_checks_policy` key. According to GitHub API documentation[1], this is a required field so it needs to be defined in the configuration file (at least at this point in time).

[1]: https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#create-a-repository-ruleset


travi (approved):
thanks!

